PR #20, opened 2020-09-29
configure dependabot

dholbach:


.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+  - package-ecosystem: "npm" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "gitsubmodule"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
```
squaremo, line 17: Could/Should this be more specific?

dholbach: I stole from https://github.com/weaveworks/ignite/blob/master/.github/dependabot.yml and the example under https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#package-ecosystem

squaremo: Got it -- https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#directory advises:

> For GitHub Actions, set the directory to / to check for workflow files in .github/workflows.

squaremo (approved):
Let's give it a whirl!

Issue #33, opened 2020-10-30
Spacing between elements require a slight change


hiddeco:
While I was mocking the structure of the homepage, I did not spend much time on padding and/or margin between elements. This is most obvious in the "features" section, where the spacing between icons and/or text is not optimal.

PR #37, opened 2020-11-04
Various tweaks

### LINKED ISSUE #29: Change gradient
Gradient of the hero elements should be changed to the one that is currently in use for the menu on https://docs.fluxcd.io, as it is a brighter / more vivid color combination.

```
background-image: linear-gradient(45deg, rgb(0, 150, 225) 0%, rgb(27, 141, 226) 24%, rgb(42, 125, 227) 53%, rgb(53, 112, 227) 78%, rgb(53, 112, 227) 100%);
```

We may want to incorporate this color change in the color palette defined in https://github.com/fluxcd/fluxcd.io/blob/next-gen/assets/sass/style.sass#L12-L15

hiddeco:
Fixes #29 

Addresses part of #33

### DIFF
--- a/assets/sass/_card.sass
+++ b/assets/sass/_card.sass
@@ -11,10 +11,11 @@
         h1, h2, h3, h4, h5, h6
           color: $white
 
-  &.equal-height
-    display: flex
-    flex-direction: column
-    height: 100%
+  +tablet
+    &.equal-height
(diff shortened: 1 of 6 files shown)


dholbach (approved):
This looks great! :heart_eyes:

Issue #32, opened 2020-10-30
Rewrite content to be about Flux v2


dholbach:
@hiddeco says
> But there is other content that needs to be worked on.
> "Overview" is for Flux v1.
> Tagline is for Flux v1.

PR #38, opened 2020-11-04
Update tagline and description to be about Flux v2

dholbach:
Addresses part of: #32


hiddeco (approved):
Features items still need a rewrite as I feel they do not provide a clear coverage of "new Flux" capabilities.

PR #39, opened 2020-11-04
describe GitOps Toolkit

dholbach:
@hiddeco I feel the features section still describes Flux v2

Would you prefer a detailed discussion of features (which would likely require a bit more background...?) - WDYT?

Addresses: #32


hiddeco (approved):

hiddeco:
@dholbach it does not highlight any of the features as lead out in https://toolkit.fluxcd.io/#more-detail-on-whats-in-flux

Issue #19, opened 2020-09-29
Point to /f/community


dholbach:
Once https://github.com/fluxcd/community/issues/6 is resolved, we should point to the README from the /f/community repository as a good way into the community.

dholbach:
If this is about importing bits from various repos into the website, we might as well punt it to a later milestone.

hiddeco:
Addressed part of "linking to community" by adjusting the menu, and moved the milestone for further changes.

dholbach:
It'll be important to link to meeting information, events and get-involved kind of info very prominently.

If we fix fluxcd/community#6 and make it beautiful, can we pull that README in and e.g. make it fluxcd.io/community or /get-involved or some such?

dholbach:
It might make sense to write /f/community/README.md like a very nice guide for somebody who's new to the community ... and link off to more detailed GH docs for the nitty gritty.

We'd need to 

1) rewrite README
2) add import script (maybe re-use parts of https://github.com/kubernetes-sigs/contributor-site/blob/master/hack/gen-content.sh?)
3) work on styling of "normal pages" on the site?

dholbach:
CCing @scottrigby and @somtochiama

dholbach:
Fixed by https://github.com/fluxcd/website/pull/74

PR #42, opened 2020-11-04
Various tweaks to homepage content

hiddeco:
Addresses more of #32
Addresses part of #19

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -91,11 +91,11 @@ subtitle = "Open and extensible continuous delivery solution for Kubernetes. Pow
 [[params.flux.buttons]]
 text  = "Get started"
 color = "warning"
-url   = "#"
+url   = "https://toolkit.fluxcd.io/get-started/"
 
 [[params.flux.buttons]]
 text = "Learn more"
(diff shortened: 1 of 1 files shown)


config.toml
```diff
@@ -76,16 +82,16 @@ type  = "sans_serif"
...
 [params.flux]
 # title    = "Alternative title"
-# subtitle = "The GitOps operator for Kubernetes"
+subtitle = "Open and extensible continuous delivery solution for Kubernetes. Powered by GitOps Toolkit."
 
 [[params.flux.buttons]]
 text  = "Get started"
 color = "warning"
-url   = "#"
+url   = "https://toolkit.fluxcd.io/get-started/"
```
scottrigby, line 90: Do these need to be full URLs? If they can be relative, it may help during branch and PR previewing.

scottrigby (approved):
Just a question. Either way, looks good!

config.toml
hiddeco: They can not be relative at this moment as they point towards the Toolkit website.

PR #43, opened 2020-11-04
Reinstantiate sitewide notification

hiddeco:
But with a message aimed at guiding people towards the Flux version 1 documentation.

### DIFF
--- a/assets/sass/_notification.sass
+++ b/assets/sass/_notification.sass
@@ -1,0 +1,3 @@
+.notification
+  &.no-radius
+    border-radius: 0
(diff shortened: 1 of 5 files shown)


layouts/_default/baseof.html
```diff
@@ -9,6 +9,7 @@
     <header>
       {{ partial "navbar.html" . }}
     </header>
+    {{ partial "notification.html" }}
```
scottrigby, line 12: Looks good! But it's not site-wide. For example, not on the docs pages. Is that intended?

dholbach (approved):
LGTM!

PR #51, opened 2020-11-05
Temporarily disable equal height on cards

hiddeco:
As it does not play nicely with Safari.


scottrigby:
Looks better on Safari!

## Before:
Screen Shot 2020-11-05 at 11 47 12 AM

## After
Screen Shot 2020-11-05 at 11 47 31 AM

scottrigby (approved):

PR #52, opened 2020-11-05
Cards of equal height

hiddeco:
Follow up on #51


scottrigby (approved):
Looks good in safari! 💯

PR #53, opened 2020-11-05
Improve Netlify configuration

hiddeco:
* Update Hugo version to 0.78.0
* Create different deploy strategy for branch deploys
* Enable garbage collection, minifying and Git info
* Change base URL in Hugo config to fix navbar link


dholbach (approved):
💖

PR #54, opened 2020-11-06
Link to security document

hiddeco:
Related to #19


stefanprodan (approved):

hiddeco:
Should not be merged till I have confirmed the mailing list referred to in the document was actually created.

Issue #50, opened 2020-11-05
Write import script to pull in data from other repositories


hiddeco:
As this will become our hub for all website and documentation related things, the developer orientated `README.md` should be rewritten to welcome contributors of all kinds.

The `kubernetes/website` repository has a great example: https://github.com/kubernetes/website

dholbach:
Hum, is this https://github.com/fluxcd/community/pull/48 (addressing fluxcd/community#6)?

If yes, we could make this issue about importing docs from elsewhere, e.g. making the /f/community README be fluxcd.io/community?

dholbach:
https://github.com/kubernetes-sigs/contributor-site/blob/master/hack/gen-content.sh is what brings this kind of functionality to <https://kubernetes.dev>

cc @alisondy

dholbach:
cc @somtochiama

dholbach:
Fixed by https://github.com/fluxcd/website/pull/74

PR #74, opened 2021-01-13
Import community docs

dholbach:
This is a bit of a wild draft to get us started with #50 - let's clean it up together.


content/en/community.md
```diff
@@ -0,0 +1,99 @@
+---
```
hiddeco, line 1: Shouldn't this file be included in the `.gitignore` so that the latest version is pulled during build?

dholbach: Hum... not sure what to make of https://github.com/kubernetes-sigs/contributor-site/commit/69818fa3a79c39812044c8d095aa2eb2f3d4329f - I guess we'd need to do this in a similar fashion?

hiddeco: Means that it should be excluded, see the combination of:

https://github.com/kubernetes-sigs/contributor-site/blob/69818fa3a79c39812044c8d095aa2eb2f3d4329f/content/en/docs/.gitignore + https://github.com/kubernetes-sigs/contributor-site/blob/69818fa3a79c39812044c8d095aa2eb2f3d4329f/external-sources/kubernetes/community#L17

dholbach:
Some questions:
- Do we need to hook up the `gen-content` target in the Makefile somewhere else?
- Do we want to note down somewhere where we copied `hack/gen-content.sh` from?
- I guess we need to rework improve styling a little in `layouts/_default/single.html` - like spacing at top and bottom?

Makefile
```diff
@@ -1,16 +1,30 @@
...
+                                        fcntl.fcntl(sys.stdout, fcntl.F_SETFL, flags&~os.O_NONBLOCK);"
+
+gen-content: ## Generates content from external sources.
+	hack/gen-content.sh
+
+container-gen-content: ## Generates content from external sources within a container (equiv to gen-content).
+	$(CONTAINER_RUN) $(CONTAINER_IMAGE) hack/gen-content.sh
+
 serve:
 	hugo server \
```
hiddeco, line 14:
```suggestion
serve:
	hack/gen-content.sh
	hugo server \
```

Otherwise the pages will not be present after cloning.

hiddeco (approved):
Thank you Daniel :grapes:

dholbach:
Thanks a lot @hiddeco :strawberry:

Issue #75, opened 2021-01-19
Automate import of docs


dholbach:
Now that #74 is sorted, we might want to look into automating this. Would it be e.g. possible to have a cron job which runs the `hack/gen-content.sh` script and create a PR if there's a diff?

hiddeco:
Given we do not store a copy in this repository, I do not think it should create a PR.

Instead, we should probably run a GitHub action `x` times a day that [triggers a manual Netlify build](00340435050145502616).

dholbach:
Thanks for digging this out. This might be the best course of action.

PR #79, opened 2021-01-25
Add cronjob to deploy 'main' on netlify (every 2h)

dholbach:
Fixes: #75

### DIFF
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -1,0 +1,12 @@
+name: Re-run build from 'main' on Netlify every two hours
+on:
+  schedule:
+    - cron: '0 */6 * * *'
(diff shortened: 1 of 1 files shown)


.github/workflows/netlify.yml
```diff
@@ -0,0 +1,12 @@
+name: Re-run build from 'main' on Netlify every two hours
+on:
+  schedule:
+    - cron: '0 */2 * * *'
```
hiddeco, line 4: I would run it every 6h as we have a 1000 minute/month build limit @ Netlify.

dholbach: Good thinking - thanks!

hiddeco (approved):

Issue #77, opened 2021-01-19
Support page


dholbach:
Explain what user options are if you need help with Flux projects.

We should point out

- FAQ/Troubleshooting Docs for individual projects
- clearly what the limits of community support are
- how to submit clear information if you really found an issue
- how to buy enterprise support (this would be open to companies who offer support around Flux and its technologies)

dholbach:
In a brief discussion the following links were shared
- https://getchef.zendesk.com/hc/en-us/articles/360031184331 for writing good issues
- https://linkerd.io/enterprise/ for a tasteful example of pointing out paid support options (not being biased towards just one company, etc.)

dholbach:
Would we want the site content to live in /community or /website? I guess the former, right?

kingdonb:
We want to push people more towards the [Discussions](https://github.com/fluxcd/flux2/discussions) format since they are more long-lived, searchable, and better for permanent documentation, but personally I think with Slack support we can be more responsive and interactive, providing a better experience. I will have more thoughts about this.

I suggested using the ⏏️ protocol for FAQs that haven't made it into the discussions or documentation yet, (I am still looking for a reference that explains how the eject protocol formally works, but...) tl;dr: if a discussion belongs in the docs, adding a hot word that can notify folks like myself, or trigger automation to generate an issue somewhere. I am afraid I suspect this protocol is no longer used, as I can't find any current references to it in the handbook.

It can be less formal than I imagined, maybe we just ask someone who has posed a good question that is not in the documentation yet, to repost their question in discussions, and follow up with answers there. The hard part is getting users to check these sources if their question has already been addressed in an FAQ before posing it again in the Slack.

dholbach:
From @phillebaba ([link](https://github.com/fluxcd/community/pull/51#issuecomment-765412862):

- [ ] a big link to this page in the Fluxv2 repo
- [ ] a notice to anyone creating a new issue in fluxv2
- [ ] sticky this link in the flux Slack channel

dholbach:
All right. Let's make sure we link https://fluxcd.io/support/ from everywhere now. :sun_with_face:

PR #80, opened 2021-01-25
Add support page to website

dholbach:
Fixes: #77 

cc @kingdonb

### DIFF
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -1,11 +1,11 @@
-name: Re-run build from 'main' on Netlify every two hours
+name: Re-run build from 'main' on Netlify every six hours
 on:
   schedule:
     - cron: '0 */6 * * *'
   pull_request:
 jobs:
   build:
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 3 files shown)


.github/workflows/netlify.yml
```diff
@@ -1,11 +1,11 @@
-name: Re-run build from 'main' on Netlify every two hours
+name: Re-run build from 'main' on Netlify every six hours
```
hiddeco, line 1:
```suggestion
name: Re-run build from 'main' on Netlify
```

.github/workflows/netlify.yml
```diff
@@ -1,11 +1,11 @@
...
 on:
   schedule:
     - cron: '0 */6 * * *'
   pull_request:
```
hiddeco:
```suggestion
```

Do not think this is actually required?

You may want to change it to

```suggestion
  workflow_dispatch:
```

so that it can be triggered from the GitHub UI.

hiddeco (approved):
Super Daniel, thanks! :sun_with_face:

PR #126, opened 2021-04-06
we moved the security file to fluxcd/.github

dholbach:
New location is here: https://github.com/fluxcd/.github/blob/main/SECURITY.md

### DIFF
--- a/external-sources/fluxcd/community
+++ b/external-sources/fluxcd/community
@@ -2,4 +2,3 @@
 "/SUPPORT.md","/support.md"
 "/CONTRIBUTING.md","/contributing.md"
 "/GOVERNANCE.md","/governance.md"
-"/SECURITY.md","/security.md"
--- a/external-sources/fluxcd/.github
+++ b/external-sources/fluxcd/.github
@@ -1,0 +1,1 @@
+"/SECURITY.md","/security.md"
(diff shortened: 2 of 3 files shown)


external-sources/fluxcd/community
```diff
@@ -2,4 +2,3 @@
 "/SUPPORT.md","/support.md"
 "/CONTRIBUTING.md","/contributing.md"
 "/GOVERNANCE.md","/governance.md"
-"/SECURITY.md","/security.md"
```
stefanprodan, line 5: New line here please

stefanprodan:
First we need to remove the link from the document, no idea why SECURITY.md links to itself but it shouldn't.

dholbach:
> First we need to remove the link from the document, no idea why SECURITY.md links to itself but it shouldn't.

I could imagine it was originally placed there so if the file ever got copied around it was clear where the main copy was. I'm happy to remove the link, or point to its new location. WDYT?

stefanprodan:
@dholbach I looked at other CNCF projects, no one has such a link, for example: https://github.com/coredns/coredns/security/policy

dholbach:
> @dholbach I looked at other CNCF projects, no one has such a link, for example: https://github.com/coredns/coredns/security/policy

https://github.com/fluxcd/.github/pull/2

stefanprodan (approved):

Issue #111, opened 2021-03-19
Finish break-up of core concepts doc


dholbach:
@alisondy started work on this in #110 - looking good already!

dholbach:
@alisondy Shall we push this back until #116 is done? So we can have the global redirect working first?

dholbach:
https://github.com/alisondy/website/commit/a4b9a886dc404aa122e3d18c7acc8f4c1a26ddb3#diff-2d935a691066f5dc5402b97a3f970470419ae53a80bedf380793d96adb5c814e

Maybe get this into `f/flux2` first?

alisondy:
Sounds like a good idea 👍

kingdonb:
A PR was merged in #141, does that complete this issue?

dholbach:
No, it's not right now there's only `./docs/concepts.md` - the idea was to break this up into various docs, cf https://github.com/alisondy/website/commit/a4b9a886dc404aa122e3d18c7acc8f4c1a26ddb3#diff-2d935a691066f5dc5402b97a3f970470419ae53a80bedf380793d96adb5c814e

Neilblaze:
@dholbach can you please assign this to me? :)

PR #141, opened 2021-04-23
Various fixes

hiddeco:


dholbach:
Regarding this change:
> Index directed to `/concepts` but page was at `/core-concepts`, given
> `/concepts` is a prettier URL, I renamed the file (and thereby the URL).
If we go ahead with this (note #111 as well) we will need to tweak the redirect rules from toolkit.f.i. No strong opinion on this though. I also think it's a nicer URL!

dholbach (approved):

rishabh96b:
Looks like the redirect works correctly!

PR #148, opened 2021-04-26
Fix title of Helm use case doc

dholbach:
thanks Tamao for spotting

### DIFF
--- a/content/en/docs/use-cases/helm.md
+++ b/content/en/docs/use-cases/helm.md
@@ -1,5 +1,5 @@
 ---
-title: GitHub Actions Manifest Generation
+title: Flux for Helm Users
 linkTitle: Helm
 ---
 


stefanprodan:
Hmm so the titles are not taking from flux2 repo?

dholbach:
The frontmatter was added by me and I made a copy/paste mistake.

stefanprodan:
Why not add them to flux2, does it messes with mkdocs?

dholbach:
I don't know and there's more manual changes I needed to make than just the frontmatter. Not sure if it's worth the effort if we will move the docs in a couple of days/weeks anyway?

stefanprodan:
> Not sure if it's worth the effort if we will move the docs in a couple of days/weeks anyway?

Ok now I'm really confused, so the frontmatter will not be required inside the markdowns after we move the docs??

dholbach:
No, that's not what I meant.

The plan is to do https://github.com/fluxcd/flux2/issues/1135 , then #116  and then https://github.com/fluxcd/flux2/issues/1221. Essentially (hopefully) in a couple of days/weeks the docs will have been removed from `f/flux2` and only live in `f/website`.

The reason why there's no automatic syncing between the two repos is:
- we add frontmatter
- links to images and other assets have changed
- for some links we have to use https://gohugo.io/content-management/shortcodes/#ref-and-relref
- instead of `!!!hint` we moved to `{{% alert % }}`
- <maybe more...?>

I hope this clarifies things.

stefanprodan:
Hmm and how are we going to keep the guides and CLI docs in sync with releases? Do contributors have to open 3 PRs every time we change something in CLI? See this PR as an example: https://github.com/fluxcd/flux2/pull/1334/files

dholbach:
Why 3 PRs?

I'm a bit surprised... Frankly I had no idea there were doubts about the docs move. @hiddeco , @alisondy  and I have been discussing this for weeks and I tried to inform the Flux Dev meetings and our internal ones about everything that was going to happen.

Potentially there could be a way to rewrite docs from `f/flux2` to `f/website` but it would be a lot of work and I'm sure also quite prone to error.

dholbach:
I suggest we discuss this on the mailing list and maybe just approve this PR for now?

stefanprodan (approved):

Issue #154, opened 2021-04-26
Fix import of components docs


dholbach:
@stefanprodan says:
> list all released tags, order by semver, then use the "latest" tag to pull docs

PR #155, opened 2021-04-26
Fix list of controller tags, version sort

dholbach:
Fixes: #154

### DIFF
--- a/hack/import-flux2-docs.sh
+++ b/hack/import-flux2-docs.sh
@@ -8,7 +8,7 @@ controller_version() {
     echo "Please install 'jq'."
     exit 1
   fi
-  curl -s "https://registry.hub.docker.com/v2/repositories/fluxcd/$1/tags/?page_size=1" | jq -r .results[].name
+  curl -s "https://api.github.com/repos/fluxcd/$1/releases" | jq -r '.[] | .tag_name' | sort -V | tail -n 1
 }
 
 gen_crd_doc() {


hack/import-flux2-docs.sh
```diff
@@ -8,7 +8,7 @@ controller_version() {
     echo "Please install 'jq'."
     exit 1
   fi
-  curl -s "https://registry.hub.docker.com/v2/repositories/fluxcd/$1/tags/?page_size=1" | jq -r .results[].name
+  curl -s "https://api.github.com/repos/fluxcd/$1/releases" | jq -r '.[] | .tag_name' | sort -V | tail -n 1
```
stefanprodan, line 11: Does this work?

```
$  curl -s https://api.github.com/repos/fluxcd/kustomize-controller/releases/latest
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/repos#get-the-latest-release"
}
```

stefanprodan (approved):

hack/import-flux2-docs.sh
hiddeco, line 11: This will give unexpected results with prereleases, see:

```console
$ printf '%s\n' v1.2.0 v.1.1.0 v.1.2.0-alpha v1.3.0-test v2.0.0-alpha | sort -V | tail -n 1
v.1.2.0-alpha
```

PR #167, opened 2021-04-28
Update helmreleases.md

Jean-Baptiste-Lasselle:
Hi, I believe, at https://fluxcd.io/docs/guides/helmreleases/#cloud-storage , I believe that you have a typo:  I think that it's a `FluxCD` `HelmRelease`, to helm release Chartmuseum, not a `HelmRepository`, see proposed change

Thank you very much fo he work done on the 2nd generation of `FluxCD` (btw, I did not find where the "gitops engine" common project with Argo CD is... any link ?)

### DIFF
--- a/content/en/docs/guides/helmreleases.md
+++ b/content/en/docs/guides/helmreleases.md
@@ -138,7 +138,7 @@ for details about how to use other storage backends.
 
 ```yaml
 apiVersion: source.toolkit.fluxcd.io/v1beta1
-kind: HelmRepository
+kind: HelmRelease
 metadata:
   name: chartmuseum
   namespace: flux-system


Jean-Baptiste-Lasselle:
my msitagke, I jsut missed that the helm repository was the helm repository for the ChartMuseum Helm Chart, and then in same Ymal the Healm Release for Chart Museum

dholbach:
> Thank you very much fo he work done on the 2nd generation of `FluxCD` (btw, I did not find where the "gitops engine" common project with Argo CD is... any link ?)

https://fluxcd.io/docs/migration/faq-migration/#is-the-gitops-toolkit-related-to-the-gitops-engine

Jean-Baptiste-Lasselle:
Hi @dholbach ,

A word to seriously thank you for letting me know about "FluxCD and Gitops Engine, where are we?". That is very important information : 
* gitops engine has incompatibility with FluxCDF v2
* Now this is , tchnically, very important information : This means that at a conceptual level, there are divergence between Gitops Engine, and FluxCD
* In other words, FluxCD and Gitops Engine see gitops differently, think of gitops with different, incompatible concepts : Or the very interesting question here is "Do  they actually ? Is there not a set of concepts that could actually unify the two apparently incompatible approaches?"

This is why I am glad that the FluxCD project shows strength, and advocates its own concepts : It allows us to raise and think of higher level idea, because it will take a very powerful idea, to reconcile Gitops Engine (let's say it, ArgoCD), and FluxCD.

Thank you FluxCD Team / Stefan Prodan for your amazing opinionated work, I love FluxCD :)

Issue #169, opened 2021-04-28
components: some pages are not listed in the nav


dholbach:
Connected to #124 we need to generate frontmatter more reliably. 

One comment from Slack:
> whole pages are missing like https://github.com/fluxcd/image-reflector-controller/blob/main/docs/spec/v1alpha2/imagerepositories.md
> https://github.com/fluxcd/image-reflector-controller/blob/main/docs/spec/v1alpha2/imagepolicies.md
> https://github.com/fluxcd/image-automation-controller/blob/main/docs/spec/v1alpha2/imageupdateautomations.md

Issue #124, opened 2021-03-26
Remove duplicate titles for imported docs


dholbach:
`hack/import-flux2-docs.sh` adds frontmatter for both imported markdown and html docs.

Now it should remove the title after having added frontmatter title entry. A good example for this now is `/docs/components/source/gitrepositories/`.

dholbach:
https://github.com/dholbach/website/blob/juozas-changes/hack/import-flux2-docs.sh#L14

PR #177, opened 2021-04-28
Fix duplicate titles in components docs

dholbach:
Make script more explicit too.

Fixes: #124
Fix: #169


dholbach:
Will likely fix #169 too.

stefanprodan (approved):
Thanks @dholbach this fixes both issues.

hiddeco (approved):

Issue #175, opened 2021-04-28
hack/gen-content.sh breaks on Mac


dholbach:
```
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
make: *** [gen-content] Error 1
```

@stefanprodan Can you please run the script using `bash -x` and paste the output here?

stefanprodan:
I think a better alternative would be to have a `make docker-preview` and run all of this inside a container, os that it works on Windows too.

dholbach:
I'm sure we could lift some of the bits from https://github.com/kubernetes/contributor-site/blob/master/Makefile

dholbach:
I think it'd still be good to have the script work on both platforms, so if you get a chance to run it under `bash -x` and paste the output here, that'd be great and I'd make sure to file the issue "upstream" as well.

stefanprodan:
```
website $ bash -x ./hack/gen-content.sh
+ set -o errexit
+ set -o nounset
+ set -o pipefail
+ readonly DEBUG=false
+ DEBUG=false
++ git rev-parse --show-toplevel
+ readonly REPO_ROOT=/Users/stefanprodan/go/src/github.com/fluxcd/website
+ REPO_ROOT=/Users/stefanprodan/go/src/github.com/fluxcd/website
+ readonly CONTENT_DIR=/Users/stefanprodan/go/src/github.com/fluxcd/website/content/en
+ CONTENT_DIR=/Users/stefanprodan/go/src/github.com/fluxcd/website/content/en
+ readonly TEMP_DIR=/Users/stefanprodan/go/src/github.com/fluxcd/website/_tmp
+ TEMP_DIR=/Users/stefanprodan/go/src/github.com/fluxcd/website/_tmp
[...]

PR #183, opened 2021-04-28
make docker-preview (okteto up)

kingdonb:
In order to support a local preview environment in a container, in light of #175, we decided to start by building a hugo container image with the right version of any other supporting software that is required to run the website locally.

This is `website:hugo-support` which I imagine we would push as `fluxcd/website:hugo-support` or maybe something with a version number in it. Then people don't have to build anything, they can just run `make docker-preview` and it will work.

There is an issue with `make serve` though preventing `make docker-preview` from working how we imagined. It works fine if you change `make serve` to run hugo with the bind address as something other than 127.0.0.1, but listening on localhost in the docker container doesn't mesh with `docker run [image] -p` which expects you to make your process as listening on 0.0.0.0, or whatever public ip.

We can talk about how to resolve it, I don't have strong opinions, as long as we publish the image `fluxcd/website:hugo-support` then contributors can just run `okteto up` on Windows, Mac, or Linux, and start making changes and start contributing to the docs right away, no docker needed at all on their machines. (They will need a Kubernetes cluster though, but it doesn't have any special requirements. It can even be an empty Kubernetes cluster with no other services installed.)

For people that may have docker installed and it's actually a lower barrier to entry than asking them to get a Kubernetes cluster, we should of course make sure that `make docker-preview` works as well. I just didn't want to change `make serve` without talking to anyone. I will come back to this with more improvements after discussion, or we can merge it and iterate.


docker-support/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM fluxcd/website:hugo-0.78.2-extended
```
stefanprodan, line 1: Where is the Dockerfile for this image?

okteto.yml
```diff
@@ -0,0 +1,13 @@
+name: hugo
+image: fluxcd/website:hugo-support
```
stefanprodan, line 2: How are users suppose to use this? There is no such image on Docker Hub.

docker-support/Dockerfile
kingdonb: It's in the `hugo` directory, which is a clone at the particular version we need

See this part of the Makefile:
```
hugo:
        git clone https://github.com/gohugoio/hugo.git -b v$(HUGO_VERSION)

.PHONY: docker-support
docker-support: hugo
        docker build -t $(DEV_IMAGE_NAME) --build-arg HUGO_BUILD_TAGS=extended hugo/
```

We would not normally need to do this, but Hugo doesn't publish their own images, with or without the `HUGO_BUILD_TAGS=extended` arg set. (I believe I got the specific instructions about how Hugo had to be built from `@dholbach`, via `@juozasg`)

The image is built by the `docker-support` target, I don't think it needs to change very often, I envisioned that one of us would build it and push it once, and there is no immediate need for automation around this part.

okteto.yml
kingdonb: We will have to publish it there. It does not need to change often, but if we think it needs more scaffolding so this image can be kept up to date, ...

docker-support/Dockerfile
kingdonb: When I tested this, all the images were of the format `kingdonb/fluxcd-website:hugo-version` etc. I figured that wasn't appropriate to publish in the fluxcd org.

I changed it here, but someone with access to the fluxcd credentials will have to build and publish these two support images just once before this can be merged.

stefanprodan: I can build it and push it to Docker Hub if you want.

kingdonb:
This PR should have included some instructions to the maintainer.

Anyone that wants to test this on your own can set your Docker Hub name in place of `fluxcd` anywhere eg. okteto.yml, in the Makefile, just put your username in `DEV_IMAGE_REGISTRY_NAME` in the makefile, set to your own DockerHub handle that you are logged in. Run `make docker-image` <- which runs the `docker-support` target. Two targets are needed simply because Hugo does not publish their own releases as images, apparently.

If you have those credentials and want to populate the images, you can just run `make docker-push` now and it will build and push the images for docker support (hugo core and the specific needs on top based on scaffolding that is in our website).

stealthybox:
Suggestion for the docker bind address thing:

```Makefile
HUGO_BIND_ADDRESS ?= 127.0.0.1

serve: gen-content theme ## Spawns a development server.
	hugo server \
		--bind=$(HUGO_BIND_ADDRESS) \
		--buildDrafts \
		--buildFuture \
		--disableFastRender

.PHONY: docker-make-serve
docker-make-serve:
	docker run -v $(shell pwd):/site -p 1313:1313 -it $(PREVIEW_IMAGE_NAME) \
		make serve HUGO_BIND_ADDRESS=0.0.0.0

```

Makefile
```diff
@@ -76,3 +83,30 @@ branch-build: gen-content theme ## Builds a Git branch (for e.g. development bra
...
+docker-image: docker-support
+	docker build -t $(PREVIEW_IMAGE_NAME) docker-support/
+
+.PHONY: docker-make-theme
+docker-make-theme:
+	docker run -v $(shell pwd):/site -it $(PREVIEW_IMAGE_NAME) make theme
+
+.PHONY: docker-make-serve
+docker-make-serve:
+	docker run -v $(shell pwd):/site -p 1313:1313 -it $(PREVIEW_IMAGE_NAME) make serve
```
stealthybox, line 104: When you nest make within docker, you'll probably want to do
```Makefile
$(MAKE) \"MAKEFLAGS=$(MAKEFLAGS)\" serve
```
instead of
```Makefile
make serve
```

example: https://github.com/weaveworks/ignite/blob/2f81d05/Makefile#L100

kingdonb: I think `$(MAKE)` is the wrong idea, I got

```
$ make docker-serve
docker run -v /Users/kingdonb/website:/site -p 1313:1313 -it fluxcd/website:hugo-support \
		/Library/Developer/CommandLineTools/usr/bin/make \"MAKEFLAGS=\" serve BIND_ADDRESS=0.0.0.0
docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "/Library/Developer/CommandLineTools/usr/bin/make": stat /Library/Developer/CommandLineTools/usr/bin/make: no such file or directory: unknown.
ERRO[0004] error waiting for container: context canceled
make: *** [docker-serve] Error 127
```

I don't think we need to pass in MAKEFLAGS or use the MAKE from the host, not sure why that would be needed, but I included the parts of what I think you suggested that didn't break as shown above.

I'm giving this one last test down each path before I'm ready to request a final review.

kingdonb (commented):
I think this is ready for review and merge now.

docker-support/Dockerfile
kingdonb: I added this information to the README.

dholbach: Nice, can we push the image to make it easier to try this out?

dholbach: :+1:

dholbach (approved):
nice work Kingdon!

Issue #129, opened 2021-04-16
Create adopters page for Flux 2 users 


dholbach:
Create adopters page for Flux 2 users.

dholbach:
It'd be great if we could have a `ADOPTERS.yml` (it could live in `f/flux2` as well?) where folks can add company name and url to logo and we auto-generate the page? Not sure if that's too much effort though...

scottrigby:
💯 We could set up an issue template for this as well

dholbach:
> 100 We could set up an issue template for this as well

Ah, so instead of people adding an entry to the yaml, they'd file an issue for this? I think I'd prefer the PR approach, as it'll be easier to spot who added which entry?

scottrigby:
@dholbach Yep agreed, ultimately this is what I meant 😸  https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

dholbach:
ah yes!

Issue #176, opened 2021-04-28
Add tab(s) shortcodes


dholbach:
The `tab` and `tabs` bits from https://github.com/tektoncd/website/tree/main/layouts/shortcodes would be nice to have, so we could do something like https://raw.githubusercontent.com/tektoncd/cli/main/docs/README.md

dholbach:
Some of the CSS magic would be needed too.

alisondy:
https://www.docsy.dev/docs/adding-content/shortcodes/#tabbed-panes

they're already built in

dholbach:
@stefanprodan Where would you like to see this used? In https://fluxcd.io/docs/get-started/#install-the-flux-cli maybe?

stefanprodan:
Lots of places, notification, install, secrets, etc everywhere we have multiple options to do the same thing.

dholbach:
Looks like we'll need a newer version of docsy though.

PR #195, opened 2021-05-03
Add adopters page and YAML config

dholbach:
For Flux v1 and Flux v2, driven by yaml files.

Fixes: #129 

~~Needs #176 to be landed first.~~


dholbach:
Needs https://mikefarah.github.io/yq/ ...

dholbach:
This is not very sophisticated yet, but see https://github.com/dholbach/website/blob/fix-129/adopters/flux-v2.yaml from which https://609019180b1d6800085b92f0--fluxcd.netlify.app/adopters/flux-v2/ is generated. It uses a modified form of https://www.docsy.dev/docs/adding-content/shortcodes/#card-panes which we get through #176 .

It'd be nice to simplify https://github.com/dholbach/website/blob/fix-129/Makefile#L49-L51 - it's required for the Netlify build.

dholbach:
How do you feel about https://github.com/dholbach/website/tree/fix-129/adopters#readme - is this too complicated?

dholbach:
I feel like this is getting somewhere. Things to be figured out still:

- [ ] Page description not working yet
- [ ] Size of logos
- [ ] No checking of paths
- [ ] carousel on main page

Do we already want to advertise the pages even if we haven't added companies yet?

adopters/flux-v2.yaml
```diff
@@ -0,0 +1,15 @@
...
+  url: /adopters/flux-v2
+  description: >
+    Organisations below all are using [Flux v2](https://fluxcd.io) in production.
+      
+    We are happy and proud to have you all as part of our community! 💖
+  companies:
+    - name: Weaveworks
+      url: https://weave.works
+      logo: /img/logos/weaveworks.png
+    - name: Xenit SE
```
phillebaba, line 12: Could you changes this to just Xenit?

Makefile
```diff
@@ -46,6 +46,10 @@ $(FONT_AWESOME_TARGET): ## Downloads the Docsy Font Awesome dependency.
 gen-content: ## Generates content from external sources.
 	hack/gen-content.sh
 	hack/import-flux2-assets.sh
+	command -v yq > /dev/null || test -x ./yq || \
+		curl -Lfs https://github.com/mikefarah/yq/releases/download/v4.7.1/yq_linux_amd64 -o yq && \
+		chmod a+x yq
```
dholbach, line 51: This is not very nice. Can anyone help find a solution which works on the Netlify builds? (Ubuntu AFAIK.)

config.toml
```diff
@@ -106,6 +106,18 @@ parent = "Project"
...
+parent = "Project"
+url = "/adopters/flux-v1"
+weight = 80
+
+[[menus.main]]
+name = "Flux v2 Adopters"
+parent = "Project"
+url = "/adopters/flux-v2"
+weight = 80
+
```
dholbach, line 120: Maybe let's not advertise these pages yet?

dholbach:
Let's move the main-page carousel to a separate PR.

dholbach:
From now onwards, I'll need help making this pretty. :grin: :art: :sparkles:

adopters/flux-v1.yaml
```diff
@@ -0,0 +1,9 @@
+adopters:
+  project: Flux v1
+  url: /adopters/flux-v1
```
dholbach, line 3: Is this the URL we want to use?

dholbach:
Cards are the same size, I need help centering the logos now. :grin:

dholbach:
Using card-deck with 5 entries per row. Fixes same size problem. Maybe good enough for now?

Makefile
stefanprodan: I see no issues with this, we do the same thing with GH Actions: https://github.com/fluxcd/flux2-kustomize-helm-example/blob/main/.github/actions/tools/entrypoint.sh

scottrigby:
Minus adding logos for everyone, the adopters page looks good in general!

I do see something strange on a small viewport – on iPhone in portrait mode, cards are in a single line. In landscape mode, they alternate between rows of 3 and rows of 2. This could be iterated on though, as IMO it looks pretty good to start.

One thing – do we want to add some direct link to this adopters page from the main fluxcd.io landing page as part of this PR?

dholbach:
Thanks for the feedback!

> Minus adding logos for everyone, the adopters page looks good in general!

Once it's up, we'll ask folks to add their logos.


> I do see something strange on a small viewport – on iPhone in portrait mode, cards are in a single line. In landscape mode, they alternate between rows of 3 and rows of 2. This could be iterated on though, as IMO it looks pretty good to start.

Yes, agreed - this is a known issue I'll need help with. Getting this into the shape it is was already a quite a bit of work (for me :innocent: )...


> One thing – do we want to add some direct link to this adopters page from the main fluxcd.io landing page as part of this PR?

Yes, if we feel it's good enough, I'm happy to add it. In the **Project** dropdown?

dholbach:
I added it to the navbar under **Project**.

stefanprodan (approved):
LGTM

Thanks @dholbach 🏅

PR #223, opened 2021-05-06
Add Luminovo to adopters

timonbimon:
We went into production this week with flux2 and flagger. 🙌
Thanks a lot for the awesome tool.

Also added .DS_Store to the .gitignore since that caught me off-guard. 😬


dholbach (approved):
Thanks a lot @timonbimon - thanks for the PR and welcome to the community!

Issue #196, opened 2021-05-04
Review search functionality


dholbach:
Docsy gives us this: https://www.docsy.dev/docs/adding-content/navigation/#site-search-options

We should take a look and figure out what works best for us.

stefanprodan:
Let's try algolia, Calico uses it, check it out: https://docs.projectcalico.org/about/about-calico

dholbach:
I signed up for their services. I'll let you know once they replied.

dholbach:
I got an email where they asked if I was maintainer of the site - let's see how the rest of the conversation goes... :crossed_fingers:

PR #227, opened 2021-05-07
Use Algolia search

dholbach:
Fixes: #196

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -44,7 +44,7 @@ sidebar_menu_compact = true
 #  Set to true to disable breadcrumb navigation.
 breadcrumb_disable = false
 #  Set to true to hide the sidebar search box (the top nav search box will still be displayed if search is enabled)
-sidebar_search_disable = false
+sidebar_search_disable = true
 #  Set to false if you don't want to display a logo (/assets/icons/logo.svg) in the top nav bar
 navbar_logo = true
 # Set to true to disable the About link in the site footer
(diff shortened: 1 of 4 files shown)


dholbach:
Not sure if we can test this before it's moved to `fluxcd.io`.

I followed all the steps listed in https://www.docsy.dev/docs/adding-content/navigation/#configure-algolia-docsearch - the css/js is loaded fine, but maybe it's because the config just lists fluxcd? https://github.com/algolia/docsearch-configs/blob/master/configs/fluxcd.json

Shall we merge it and roll back if it doesn't work?

dholbach:
The Algolia search is working now - MUCH better! <3

scottrigby:
Just a note to override the partial and file upstream issue. Tired kubecon convo in slack 😅

scottrigby (approved):
Looks good! only thing is, could we adjust docssearch-configs to traverse the top level above docs too, so that it picks up blogs, adopters, etc?

https://github.com/algolia/docsearch-configs/blob/5b9967e3d0124868fcd315ecb4b532c0b36d787d/configs/fluxcd.json#L4

Could be a follow up, that's my only other question so approving too 💯

dholbach:
> Looks good! only thing is, could we adjust docssearch-configs to traverse the top level above docs too, so that it picks up blogs, adopters, etc?
> 
> https://github.com/algolia/docsearch-configs/blob/5b9967e3d0124868fcd315ecb4b532c0b36d787d/configs/fluxcd.json#L4
> 
> Could be a follow up, that's my only other question so approving too 100

I sent a mail to Algolia to ask.

Issue #138, opened 2021-04-23
Blog section: review size/placement of boxes


dholbach:
Feedback we've got:
> Overview on blog page is weird, posts in 2021 (which should be of most importance) are cramped in vertical view while "posts in 2020" get full width

PR #228, opened 2021-05-07
Roll back blog list for now

dholbach:
The placement of blocks is a bit odd, move back to default blog list for now.

Fixes: #138


hiddeco (approved):
Much better (for now), thanks @dholbach 🍍

Issue #201, opened 2021-05-05
Adopters page: long company names get wrapped mid-word


dholbach:
![Screenshot from 2021-05-05 17-07-14](https://user-images.githubusercontent.com/1346979/117163989-7995fc80-adc4-11eb-9112-99c7e045ebbe.png)

juozasg:
This is fixed in https://github.com/fluxcd/website/pull/230 and should be closed? @dholbach

dholbach:
Unfortunately not fixed in Firefox yet.

juozasg:
It seems related to Firefox refused to hyphenate capitalized words. For example, this does work in Firefox with no changes to CSS:

![image](https://user-images.githubusercontent.com/4100/124201851-502ce080-da96-11eb-8717-35d555fe438a.png)

It is possible to fix this by inserting Soft Hyphen hints `&shy;` in CamelCase words. I've done this in flux and flagger adopters .yaml page for `Media&shy;Markt&shy;Saturn` and `UK Hydro&shy;graphic Office`  (previously it hyphenated `Hydrographi-c`) and it seems to look ok in Firefox now. I'm not seeing any other hyphenation problems in the page but could be missing some. If not, this can be closed I think?

PR #230, opened 2021-05-07
Add hyphenation

dholbach:
This fixes the situation on Chromium at least.

Screenshot from 2021-05-07 15-09-23


dholbach:
Follow up to #201

stefanprodan (approved):

PR #232, opened 2021-05-09
Fix CA ref in helm-operator-migration

christiangonre:
Fixed name of caCert key use as secretRef for a private HelmRepository, the right name of the key is caFile.


christiangonre:
Done!

stefanprodan (approved):
LGTM

Thanks @christiangonre

PR #233, opened 2021-05-10
Add page description to docs

stefanprodan:


hiddeco:
Think that the reason they do not render nicely is because they are "code tabs", and assumed automatically to contain code (with in every tab an example for another code language).

hiddeco:
They have added some additional snippets to make this possible, see: https://github.com/tektoncd/website/tree/main/layouts/shortcodes

stefanprodan:
@hiddeco  I added the partials and their CSS and now nothing is displayed...

hiddeco:
There is a missing piece of Javascript that takes care of showing and hiding the elements: https://github.com/tektoncd/website/blob/9c85dfb5ab6f1f63824186448142cc8e0c3d0b4c/assets/js/tabs.js

I also think the Tekton project should be credited in the files, as the work seems to be original.

stefanprodan:
@hiddeco let's merge this, I gave up on using tabs

hiddeco (approved):

PR #234, opened 2021-05-10
Add favicons

stefanprodan:
Add favicons to trim down the 404s

Screenshot 2021-05-10 at 17 13 30


hiddeco:
I do not think that this pull request actually solves the issue, except for tricking the Netlify statistics, as the `head` element still contains:

```html
<link rel="shortcut icon" href="/favicons/favicon.ico">
<!-- and other formats -->
```

PR #261, opened 2021-05-13
Improved hero banner design

juozasg:
Please merge this gradient if you can. It contains a CSS3 linear repeating gradient at 15 degree angle that matches your previous design idea. It also contains two radial gradients that gives a 3D look. I think it looks good but obviously that is subjective. 

I also added an hacky bash (sh) alert to hack/import-flux2-assets.sh that make sure the script will fail with correct warning if you do not have jq installed.


content/en/_index.md
```diff
@@ -0,0 +1,219 @@
+---
```
dholbach, line 1: This PR seems to re-add `content/en/_index.md`. I moved over over to `content/en/_index.html` as we added a lot of HTML in the landing page and upon using a newer `docsy` version our generated HTML wasn't valid any more, etc. We can't have `_index.md` and `_index.html` for generating the same page.

dholbach:
Thanks for all the other changes - this looks a lot cleaner. I like it very much! <3

dholbach (changes requested):
If you remove `_index.md` we should be ready to merge this!

content/en/_index.md
juozasg: On juozasg:main branch I have deleted _index.md that I accidentally committed.

dholbach (approved):
Nice work @juozasg !

PR #264, opened 2021-05-14
Fix the link that goes to the installation guide

mmorejon:


dholbach (approved):
Thanks a lot for the fix @mmorejon !

PR #267, opened 2021-05-17
Fixup docker support

yebyen:
Add coreutils to `docker-support/Dockerfile`

* some other changes related to making it a bit easier for maintainers to keep this up to date.

If a maintainer can run `make docker-push-support` that is the quickest way to bring the image current, when this gets merged.

The errors this fixes are related to recent changes in support scripts:

    * tail --lines=+2 in hack/import-flux2-assets.sh

(Fuller/more complete details are in the commit message)

PR #274, opened 2021-05-19
Jenkins + Flux guide

kingdonb:
This is the Jenkins+Flux use case guide that I've been promising to deliver

- [x] Verify all the links are correct
- [x] Test the Jenkinsfile as described actually works
- [x] Prepare the remote example repo that demonstrates it works

This works for me - any type of feedback is welcome! RFC

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,11 +15,13 @@ FONT_AWESOME_TARGET        := themes/$(DOCSY_COMMIT_FOLDER)/assets/vendor/$(FONT
 
 DEV_IMAGE_REGISTRY_NAME    := fluxcd
 HUGO_VERSION               ?= 0.78.2
-DEV_IMAGE_BASE_NAME        := website:hugo-$(HUGO_VERSION)-extended
-PREVIEW_IMAGE_BASE_NAME    := website:hugo-support
-DEV_IMAGE_NAME             := $(DEV_IMAGE_REGISTRY_NAME)/$(DEV_IMAGE_BASE_NAME)
-PREVIEW_IMAGE_NAME         := $(DEV_IMAGE_REGISTRY_NAME)/$(PREVIEW_IMAGE_BASE_NAME)
+HUGO_IMAGE_BASE_NAME       := website:hugo-$(HUGO_VERSION)-extended
+SUPPORT_IMAGE_BASE_NAME    := website:hugo-support
+HUGO_IMAGE_NAME            := $(DEV_IMAGE_REGISTRY_NAME)/$(HUGO_IMAGE_BASE_NAME)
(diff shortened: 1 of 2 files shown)


kingdonb:
This PR is included in #267 now, to simplify things a bit for me. We can merge both at the same time, but some changes from there crept into this PR necessarily, so it will be harder to disentangle them now.

(The `make docker-preview` functionality is currently a tad bit broken until #267 is merged. The script which looks for "python" does not specifically look for python3, but in alpine it seems you must be more specific, or add a symlink.

https://github.com/fluxcd/website/pull/267/files#diff-d5392de4f8ffaff5c05a70db2626d62901642d3fee8336f0bfab9b5531fb9cd3R19 <- the fix)

kingdonb:
This extremely topical video just came out, I am thinking of adding it as a reference to explain "what do you mean by traditional CI/CD pipeline"

https://www.youtube.com/watch?v=2Kc3fUJANAc

kingdonb:
Things I think are missing:

* No help to install Jenkins (maybe we should provide some guidance here)
* No example ImagePolicy or ImageUpdateAutomation resources
* No guidance about how to configure Jenkins in a more declarative way (JCasc?)

There are enough external references that I think this could merge as-is, some of the above things are potentially "kubecon talk material" that are maybe too much for our docs. If we think adding any of the above additionally would be helpful, I can churn out quickly some additional text and examples.

kingdonb:
I'm happy if this gets merged with #267, if there are any concerns we can address them in a follow-up PR later.

kingdonb:
(Comments rehashing issues where this PR is possibly a little big too light)

We could provide some guidance about the best way to install Jenkins, but what I've found from working with actual Jenkins users is they typically have a machine set aside for Jenkins, often it does not run Kubernetes, and basically this is fine and hardly even changes the content of the example at all; we don't show how to install and configure the Kubernetes plugin because it is basically taken care of for you if your installation of Jenkins was done with the Jenkins-CI official helm chart.

It would be good to acknowledge these users and point to docs that cover the Kubernetes plugin, so that the test stage in Jenkinsfile which invokes `agent kubernetes` will work. The rest of Jenkinsfile is simpler with a non-Kubernetes Jenkins host.

I personally choose to run Jenkins on Kubernetes this way which simplifies a few things but invites some new complications too.

Setting up Jenkins with Flux and doing declarative everything, including webhooks, dns, etc. would be a good addition for future expansion on this tutorial, might make it more accessible to people that might not have been Jenkins users before.

dholbach (approved):
I can't judge the actual technical content, but it's well-written and generally LGTM!

PR #275, opened 2021-05-19
fix(doc): watch command

hervelemeur:


stefanprodan:
See https://en.wikipedia.org/wiki/Watch_(command)

PR #299, opened 2021-05-27
Ways of structuring your repositories

stefanprodan:
This PR adds a guide to showcase several approaches of organizing repositories for a smooth GitOps experience with Flux.

Preview: https://deploy-preview-299--fluxcd.netlify.app/docs/guides/repository-structure

### DIFF
--- a/content/en/docs/guides/repository-structure.md
+++ b/content/en/docs/guides/repository-structure.md
@@ -1,0 +1,246 @@
+---
+title: "Ways of structuring your repositories"
+linkTitle: "Ways of structuring your repositories"
+description: "How to structure your Git repositories for a smooth GitOps experience with Flux."
+weight: 10
+---
+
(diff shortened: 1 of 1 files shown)


content/en/docs/guides/repository-org.md
```diff
@@ -0,0 +1,246 @@
...
+    ├── production 
+    └── staging
+```
+
+A complete example of this approach can be found at
+[flux2-multi-tenancy](https://github.com/fluxcd/flux2-multi-tenancy).
+
+### Delivery
+
+The delivery process is similar to the [monorepo](#delivery-management) one. The main difference is the
```
hiddeco, line 124:
```suggestion
The delivery process is similar to the [monorepo](#monorepo) one. The main difference is the
```

stefanprodan: The link points to the monorepo delivery management section

PR #300, opened 2021-05-27
make adopters screen responsive based on screen width

Frederik-Baetens:
I thought this looked nicer when the website is half width or less on my screen.


dholbach (approved):
Indeed - this looks great - thanks a lot @Frederik-Baetens !

PR #310, opened 2021-06-02
Add flux v1 and helm-operator docs

dholbach:
Let's move our v1 docs to fluxcd.io as well to give us more control over the docs again, simplify our varied setups, make it clearer to users what they're looking at and eventually free up docs.fluxcd.io again.

One drawback will be: we won't have versioned deploys like we do on docs.fluxcd.io right now.

In this PR I bring our overridden `layouts/partial.html` closer to `docsy` again - we effectively dropped the version selector in our fork. New diff to upstream navbar looks like so (we had a much bigger diff):

```diff
--- themes/docsy/layouts/partials/navbar.html	2021-06-02 12:13:03.123353026 +0200
+++ layouts/partials/navbar.html	2021-06-02 14:08:28.179406624 +0200
@@ -3,7 +3,6 @@
         <a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
 		<span class="navbar-logo">{{ if .Site.Params.ui.navbar_logo }}{{ with resources.Get "icons/logo.svg" }}{{ ( . | minify).Content | safeHTML }}{{ end }}{{ end }}</span><span class="text-uppercase font-weight-bold">{{ .Site.Title }}</span>
 	</a>
-	<div class="td-navbar-nav-scroll ml-md-auto" id="main_navbar">
 		<ul class="navbar-nav mt-2 mt-lg-0">
 			{{ $p := . }}
 			{{ range .Site.Menus.main }}
@@ -16,20 +15,33 @@
 				{{ $post := .Post }}
 				{{ $url := urls.Parse .URL }}
 				{{ $baseurl := urls.Parse $.Site.Params.Baseurl }}
+				{{ if .HasChildren }}
+				<li class="nav-item dropdown">
+					<a class="nav-link {{if $active }} active{{end}} dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+						{{ .Name }}
+					</a>
+					<div class="dropdown-menu dropdown-menu-right" aria-labelledby="navbarDropdownMenuLink">
+					{{ range .Children  }}
+						{{ $childrenurl := urls.Parse .URL }}
+						<a class="dropdown-item" href="{{ with .Page }}{{ .RelPermalink }}{{ else }}{{ .URL | relLangURL }}{{ end }}" {{ if ne $childrenurl.Host $baseurl.Host }}target="_blank" {{ end }}>{{ .Name }}</a>
+					{{ end }}
+					</div>
+				</li>
+				{{ else }}
 				<a class="nav-link{{if $active }} active{{end}}" href="{{ with .Page }}{{ .RelPermalink }}{{ else }}{{ .URL | relLangURL }}{{ end }}" {{ if ne $url.Host $baseurl.Host }}target="_blank" {{ end }}>{{ with .Pre}}{{ $pre }}{{ end }}<span{{if $active }} class="active"{{end}}>{{ .Name }}</span>{{ with .Post}}{{ $post }}{{ end }}</a>
+				{{ end }}
 			</li>
 			{{ end }}
 			{{ if  .Site.Params.versions }}
-			<li class="nav-item dropdown d-none d-lg-block">
+			<li class="nav-item dropdown d-lg-block">
 				{{ partial "navbar-version-selector.html" . }}
 			</li>
 			{{ end }}
 			{{ if  (gt (len .Site.Home.Translations) 0) }}
-			<li class="nav-item dropdown d-none d-lg-block">
+			<li class="nav-item dropdown d-lg-block">
 				{{ partial "navbar-lang-selector.html" . }}
 			</li>
 			{{ end }}
+			<li class="nav-item nav-search-item">{{ partial "search-input.html" . }}</div>
 		</ul>
-	</div>
-	<div class="navbar-nav d-none d-lg-block">{{ partial "search-input.html" . }}</div>
 </nav>

```


dholbach:
see e.g. https://deploy-preview-310--fluxcd.netlify.app/legacy/flux/faq/ and the top version picker

dholbach:
Making it it's own `type`, we can style it however we want, e.g. use different background colour, or however we like.

Everything would live under `fluxcd.io/legacy/flux` (and `helm-operator`) with no changes to the URL structure, so redirects from docs.fluxcd.io should be fairly straight-forward. Please note that this won't give us `latest` and `stable` and `1.22.0` versions of these docs.

In a step further down the road (once our legacy projects are phased out), we could make `docs.fluxcd.io` redirect to `fluxcd.io/docs`.

dholbach:
Things known to be broken:
- `!!!hint`
- https://github.com/dholbach/website/blob/add-flux-v1-docs/layouts/legacy/baseof.html#L26-L29 needs better text and maybe more beauty
- diagram is huge
- order of things in nav (weight)
- probably more

kingdonb:
### diagram is huge
I was able to make a change that fixes the main diagram so it is not huge, I will suggest it as an addition to the PR...

I added some CSS which simply sets it to 100% width, which looks fine to me. There might be a more elegant or consistent way with how you are handling CSS to resolve this.

If diagrams are high pixel count, they each need this type of formatting applied to them (width=100%). I will come back and update this checklist after first pass then go through every page in the v1 docs, looking for anywhere images are used, to note any similar issues, and document them here:

- [X] Main diagram at landing
- [ ] ...
- [ ] TODO: list more diagrams here so they can be fixed together

### code blocks are not properly formatted

I found that code blocks are missing the 10px padding format, and rounded corner bevel/border formatting.

They also had a "copy" control which appears to be missing on all of the migrated v1 docs wherever pre-formatted code blocks are used.

Example:

Screen Shot 2021-06-02 at 3 09 42 PM

This formatting maybe already exists and can be copied from the "current" docs CSS, I am not familiar with the structure (but would be able to figure this out with some time to spend on it.) The "current" docs do nice formatting of code blocks for sure.

Some other formatting regressions that seem like they might already be addressed already in "current" docs, or might otherwise need to be solved in both places:

### The text content is overcrowded
... and looks like there is not enough whitespace in some text places, here is an example from the guide "Upgrading to Flux v1"

New:
Screen Shot 2021-06-02 at 3 13 16 PM
Old:
Screen Shot 2021-06-02 at 3 14 59 PM

I will update this comment as I find more issues.

assets/scss/_legacy.scss
```diff
@@ -0,0 +1,6 @@
+.legacy {
+  margin-top: 6rem;
+  @include media-breakpoint-down(sm) {
+    margin-top: 2rem;
+  }
```
kingdonb, line 5:
```suggestion
  }
img[alt="Flux v1 Diagram"] {
  width: 100%;
}
```

If there is a more elegant way to solve this I'm all ears.

assets/scss/_legacy.scss
```diff
@@ -0,0 +1,6 @@
...
+}
```
kingdonb, line 6:
```suggestion
}
img[alt="Flux v1 Diagram"] {
  width: 100%;
}
```

If any image is used anywhere with this alt text, it is set to 100% width. This fixes the "diagram is huge" issue and hopefully doesn't affect anything else. (I'm not sure how to make changes that only apply to the legacy docs, if it is sufficient to put them in this file, or if there are other cases, maybe we should simply apply this formatting to every image in the legacy docs.)

content/en/legacy/flux/references/daemon.md
```diff
@@ -0,0 +1,113 @@
...
+
+2. Lock vs Unlock
+
+    Deployment is pinned to a particular image tag.
+    New deployment will not proceed upon triggered release.
+
+## Setup and configuration
+
+`fluxd` requires setup and offers customization though a multitude of flags.
+
```
kingdonb, line 40: This table is not particularly readable:

Screen Shot 2021-06-02 at 3 27 32 PM

The old look was like this:
Screen Shot 2021-06-02 at 3 28 35 PM

content/en/legacy/flux/references/garbagecollection.md
```diff
@@ -0,0 +1,48 @@
...
+you reconfigure `fluxd`. It is intended to be conservative: it ensures
+that `fluxd` will not delete resources that it did not create.
+
+## Limitations of this approach
+
+In general, if you change an element of the source (the git repo URL,
+branch, and paths), there is a possiblility that resources no longer
+present in the new source will be missed (i.e., not deleted) by
+garbage collection, and you will need to delete them by hand.
+
```
kingdonb, line 43: Similarly unreadable table

New:
Screen Shot 2021-06-02 at 3 30 40 PM

Old:
Screen Shot 2021-06-02 at 3 32 23 PM

I think this should be fixed by the same change that fixes any other table issue. I would suggest an alternate color or border-based callout look for the table header (light gray background like the old version would be fine IMHO) and a bit more whitespace in the form of padding or margin between the columns.

dholbach: maybe it's just a styling bit that's missing from the other docs...?

assets/scss/_legacy.scss
kingdonb: Looks like there is a `max-width: 100%` applied to images in the "current" docs website from somewhere. This is the more elegant way (but I don't see where it's been set from as of yet, or how to repeat it in the "legacy" docs.)

kingdonb: I think it must come from:


```
// Responsive image
//
// Keep images from scaling beyond the width of their parents.

@mixin img-fluid() {
  // Part 1: Set a maximum relative to the parent
  max-width: 100%;
  // Part 2: Override the height to auto, otherwise images will be stretched
  // when setting a width and height attribute on the img element.
  height: auto;
}
```

`themes/docsy/assets/vendor/bootstrap-4.6.0/scss/mixins/_image.scss`

If there is a simple change that will include all of the "mixins" such as this one that apparently resolves this issue in `current` docs section, that will probably correct a lot of the issues that I'm identifying with my read-through now.

dholbach: We need to make sure `type: legacy` has the same styling as `type: docs`.

dholbach:
I was able to make it use the `docs` template and just put the legacy warning in there: https://github.com/dholbach/website/blob/add-flux-v1-docs/layouts/docs/baseof.html#L26-L31

Let me know what you think.

kingdonb:
None of the issues I had highlighted up to this point are still present, from what I can tell. It looks terrific!

I'll keep reading, I found at least one bad link that it looks like was added by my last update (we had those FAQ entries that we decided to put into the Flux v1 docs FAQ)

Let me know if you need help translating `!!!hint`

If you have already a plan for dealing with the split-search, super. We will probably want to figure that out before this merges, so as to not ruin the search. This is in pretty good shape now!

dholbach:
I'm very pleased to hear that all these issues are resolved.

If you want to help with the `{{% alert %}}` bits, that's great. I will only have time post-GitOps Days for this again. Regarding the search, I'll mail the person at Algolia I last talked to, to help us set up a flux and a flux-legacy search.

Is `fluxcd.io/legacy/{flux,helm-operator}` okay as URL pattern?

content/en/legacy/_index.md
```diff
@@ -0,0 +1,4 @@
+---
+title: Flux Legacy
+type: docs
+---
```
dholbach, line 4: We might want more description on this page...

content/en/legacy/flux/requirements.txt
```diff
@@ -0,0 +1,4 @@
+mkdocs                  == 1.1.2
```
dholbach, line 1: This file can be deleted.

dholbach:
Added the helm-operator docs - some of the nav needs fixing... the `hack/gen-content.py` imports some of the files that are symlinks now - we will need to 

- [ ] adapt the script to change frontmatter to include weight and `type: docs`
- [ ] add more nav / weight
- [ ] more...?

dholbach:
Things that need doing:

- [ ] agree on URL structure (@hiddeco @kingdonb @squaremo @stefanprodan )
- [ ] break out legacy text into its own partial (Daniel)
- [ ] update and extend legacy help text (who?)
- [ ] extend `/legacy` landing page text (who?)
- [ ] look into getting two separate algolia searches (Daniel)
- [ ] general review

dholbach:
I reached out to Algolia DocSearch team to see how we can set up two searches.

dholbach:
I rewrote the legacy warning, based on what's displayed on docs.fluxcd.io - wdyt @kingdonb ?

dholbach:
Re search: until we have a second Algolia search bucket (not sure if that's what it's called), I asked for legacy docs to be excluded
 in https://github.com/algolia/docsearch-configs/pull/4190.

dholbach:
For now 'legacy' will not be indexed.

If we want to re-enable, we should take a look at https://docsearch.algolia.com/docs/config-file/#using-custom-tags it was advised - we could have split results this way.

dholbach:
@kingdonb If you could 

- [ ] review (and if necessary improve) https://github.com/dholbach/website/blob/add-flux-v1-docs/layouts/partials/docs/legacy-warning.html and
- [ ] suggest some text for https://github.com/dholbach/website/blob/add-flux-v1-docs/content/en/legacy/_index.md - maybe we can just steal from somewhere, so https://deploy-preview-310--fluxcd.netlify.app/legacy/ does not look as lonely


Then I'll 

- [ ] double-check everything is in the right place compared to our docs.fluxcd.io pieces
- [ ] update docs from f/flux and f/helm-operator if necessary one more time

Everybody please

- [ ] have another look over the site 

Re-adding the (split) search is something we can look into once this PR is landed.

Anything I've missed?

dholbach:
https://deploy-preview-310--fluxcd.netlify.app/legacy/helm-operator/get-started/using-helm/ does not seem to replace `{{ version }}`.

dholbach:
- `/legacy/helm-operator/references/helmrelease-custom-resource/` is also not quite there yet.
- `/legacy/helm-operator/references/chart/` neither

dholbach:
- `/contributing/introduction/` under helm op seems to be missing

Everything else seems to be as on docs.f.io

content/en/legacy/flux/_index.md
```diff
@@ -0,0 +1,47 @@
+---
+title: Flux documentation
```
stefanprodan, line 2:
```suggestion
title: Flux v1
```

layouts/partials/docs/legacy-warning.html
```diff
@@ -0,0 +1,10 @@
+<div class="pageinfo pageinfo-warning">
+    <h2>Legacy Flux is in Maintenance Mode</h2>
+    <p>You are looking at the documentation of Legacy Flux and Helm Operator (v1),
+       which is in maintenance mode.</p>
+    <p>The Flux community is working with most of its energy on its current inception
```
stefanprodan, line 5: Please remove this paragraph.

content/en/legacy/flux/_index.md
dholbach: I used  the term `Flux Legacy` in a couple of places - WDYT about this?

layouts/partials/docs/legacy-warning.html
dholbach: What would you like to see in this version warning? Or which lines in particular would you like to see removed?

content/en/legacy/flux/_index.md
stefanprodan: Sure let's use that instead of v1

layouts/partials/docs/legacy-warning.html
hiddeco: I would simply remove the sentence because it can cause doubt about "production readiness".

> There is a new version -> upgrade now

Really is all that should be said.

dholbach: So remove lines 5-7? WFM

stefanprodan:
```
Legacy Flux is in Maintenance Mode

This documentation is for Flux (v1) and Helm Operator (v1). Both projects are in maintenance mode and will soon reach end-of-life. We strongly recommend you familiarise yourself with the latest Flux release and start looking at your migration path.
```

kingdonb:
I had several issues on the old docs site that are already resolved in this PR, it gets a 👍 from me

Most importantly there are some pages on the old helm operator docs which still do not include the admonishment banner warning users that they are looking at a doc for an old version. From the pages I checked in the deploy preview, they are all properly admonishing now.

I've been reading through the old docs and looking for issues, I'll re-commence that effort as soon as this lands, so we don't have wasted effort cataloging issues that are already being fixed. 💯

layouts/partials/docs/legacy-warning.html
hiddeco: I would maybe rephrase it to something like:

```code
<h2>Upgrade Advisory</p>
<p>You are looking at the documentation of a version of Flux and the Helm Operator which are in maintenance mode, and have thereby become legacy projects.</p> 
<p>We strongly recommend you familiarise yourself with the newest Flux and
       <a href="/docs/migration/">start looking at your migration path</a>.</p>
```

dholbach:
Follow-up items post-merge are noted down in https://github.com/fluxcd/website/issues/327

content/en/legacy/_index.md
dholbach: Moved to https://github.com/fluxcd/website/issues/327

hiddeco (commented):
I have not reviewed all pages, but the outline looks OK.

hiddeco (approved):

PR #311, opened 2021-06-03
Update image-update.md to allow secret creation

dafstone:
Patch the current documentation to allow continuation of line where required to create the docker-registry secret in the example.  Without this the example will not run due to a bash error.  

Output of error this fixes:

```
secret/ecr-credentials created (dry run)
/bin/bash: line 5: -o: command not found
error: no objects passed to apply
```

### DIFF
--- a/content/en/docs/guides/image-update.md
+++ b/content/en/docs/guides/image-update.md
@@ -723,7 +723,7 @@ spec:
                 --dry-run=client \
                 --docker-server="$ECR_REGISTRY" \
                 --docker-username=AWS \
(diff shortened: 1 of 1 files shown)


hiddeco (approved):

hiddeco:
Thank you for the patch! Before we are able to merge we need a sign-off on your commit, https://github.com/fluxcd/website/pull/311/checks?check_run_id=2739153905 tells you how to do this.

dafstone:
@hiddeco -- all set there, thanks for the tip -- didn't realize that was facing me.  All signed.

PR #313, opened 2021-06-04
Update HelmRelease creation example

hiddeco:
...and update the example output to match what's generated with the command.

### DIFF
--- a/content/en/docs/use-cases/helm.md
+++ b/content/en/docs/use-cases/helm.md
@@ -49,7 +49,7 @@ Flux client:
 
 ```sh
 flux create source helm traefik --url https://helm.traefik.io/traefik
-flux create helmrelease --chart my-traefik \
+flux create helmrelease my-traefik --chart traefik \
   --source HelmRepository/traefik \
   --chart-version 9.18.2 \
   --namespace traefik
@@ -81,7 +81,7 @@ spec:
 apiVersion: helm.toolkit.fluxcd.io/v2beta1
 kind: HelmRelease
 metadata:
-  name: traefik
+  name: my-traefik
   namespace: traefik
 spec:
   chart:


stefanprodan:
traefik is a bad example to give as it requires a LoadBalancer and lots of resources, making this tutorial fail on anyones laptop.

hiddeco:
Forgot I had created this PR, what other chart would be a better example?

hiddeco:
I suspect @scottrigby used the example that is typically used with Helm, but I may be wrong.

stefanprodan:
@hiddeco I would use something that's validated by e2e tests, e.g. https://github.com/fluxcd/flux2-kustomize-helm-example/blob/main/infrastructure/nginx/release.yaml

dholbach:
#414 fixes this similarly...?

scottrigby:
Also changing the example chart may be better as a follow-up to this PR, which just updates the example to match CLI output?

dholbach:
Looks like #414 addresses some additional things: https://github.com/fluxcd/website/pull/414#issuecomment-888348279

scottrigby:
> Looks like #414 addresses some additional things: #414 (comment)

@dholbach I commented in that other PR: https://github.com/fluxcd/website/pull/414#issuecomment-890078841 Tl;dr, I believe this one is the correct CLI output fix/update.


> @hiddeco I would use something that's validated by e2e tests, e.g. https://github.com/fluxcd/flux2-kustomize-helm-example/blob/main/infrastructure/nginx/release.yaml

@stefanprodan I created a new issue for this that explains the benefits you noted, as well as my comment about which version of ingress-nginx chart we should use if we do: https://github.com/fluxcd/website/issues/436

scottrigby (approved):

PR #314, opened 2021-06-08
Add D2iQ to adopters

makkes:
Signed-off-by: Max Jonas Werner <mwerner@d2iq.com>


stefanprodan (approved):
LGTM

Thanks @makkes 🏅

Issue #181, opened 2021-04-28
Broken links on website


dholbach:
I turned on the check links bit in netlify, but needed to quick turn it off again as it spit out these issues:

```
...........................................................×................................××.........................................................................................×....................................................×.........................................................×××.........×....×××××.×××××.....××........................................................................................................××××..×...........................×..............................×.×..........................................................................××.............×××...×.....×××××××.×××××××××××××××××××××××××××××××××××××××××××××××××××××××.................................................................×..×××....................................................................................................................................................................................................................................................................................................................................×..............................................×......×.........................................×...................................×...............................×............................................×..................................................×.............................................................................................................×.......................................................×..........................×.......................................................................................................................................................................................................................................................................................................................................................................................××.................................................................×...............................................................×......................................................................................................................................×......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
2:51:58 PM:   ✖ FAIL load ../../../img/flagger-gitops.png
2:51:58 PM:   | operator: load
2:51:58 PM:   | expected: 200 ../../../img/flagger-gitops.png
2:51:58 PM:   |   actual: ENOENT: no such file or directory, open '/img/flagger-gitops.png'
2:51:58 PM:   |       at: public/index.xml:1172:20 (inlined Html) <img src="../../../../img/flagger-gitops.png" alt="Flagger">
2:51:58 PM:   ✖ FAIL fragment-check inline Html in public/docs/index.xml --> https://docs.google.com/document/d/1l_M0om0qUEN_NNiGgpqJ2tvsF2iioHkaARDeh6b70B0/edit
2:51:58 PM:   | operator: fragment-check
2:51:58 PM:   | expected: Fragment identifiers in links to different documents should not be empty
2:51:58 PM:   |       at: public/docs/index.xml:847:13 (inlined Html) <a href="https://docs.google.com/document/d/1l_M0om0qUEN_NNiGgpqJ2tvsF2iioHkaARDeh6b70B0/edit#" target="_blank">...</a>
2:51:58 PM:   ✖ FAIL fragment-check inline Html in public/docs/index.xml --> https://github.com/fluxcd/flux/blob/master/chart/flux/README.md
2:51:58 PM:   | operator: fragment-check
2:51:58 PM:   | expected: Fragment identifiers in links to different documents should not be empty
[...]

PR #320, opened 2021-06-11
Enable markdwon-link-check on PR

NissesSenap:
Checks for broken links in files that have been chnaged in the PR.
Update Makefile with markdown-link-check install command

### DIFF
--- a/.github/workflows/lychee-cron.yaml
+++ b/.github/workflows/lychee-cron.yaml
@@ -1,0 +1,28 @@
+name: lychee link Checker cronjob
+
+on:
+  repository_dispatch:
+  workflow_dispatch:
+  schedule:
+    - cron: "00 18 * * *"
+
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2.2.2
+        with:
+          python-version: 3.8
+      - name: "Install prerequisites"
+        run: |
+          pip3 install -r requirements.txt
+      - run: |
+          make gen-content
+      - name: Link Checker
+        uses: lycheeverse/lychee-action@v1.0.8
+        with:
+          args: --verbose --no-progress --exclude-all-private **/*.md
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
--- a/.github/workflows/lychee-pr.yaml
+++ b/.github/workflows/lychee-pr.yaml
@@ -1,0 +1,37 @@
+on: pull_request
+name: lychee link Checker
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Python
+        uses: actions/setup-python@v2.2.2
+        with:
+          python-version: 3.8
+      - name: "Install prerequisites"
+        run: |
+          pip3 install -r requirements.txt
+      - name: get changed files
+        id: getfile
+        run: |
+          echo "::set-output name=ts::$(git diff --name-only --diff-filter=ACMRT origin/main... |grep .md |xargs)"
+      - name: echo the changed files
+        run: |
+          for i in ${{ steps.getfile.outputs.ts }}
+          do
+              echo $i
+          done
+      - run: |
+          make gen-content
+      - name: lychee Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v1.0.8
+        with:
+          args: --verbose --no-progress --exclude-all-private ${{ steps.getfile.outputs.ts }}
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+      - name: Fail if there were link errors
+        run: exit ${{ steps.lychee.outputs.exit_code }}
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ SUPPORT_IMAGE_NAME         ?= $(DEV_IMAGE_REGISTRY_NAME)/$(SUPPORT_IMAGE_BASE_NA
 HUGO_BIND_ADDRESS          ?= 127.0.0.1
 BUILDER_CLI                := docker
 # BUILDER_CLI                := okteto
+LYCHEE_IMAGE_NAME          ?= lycheeverse/lychee:202105190720247e4977
 
 help:  ## Display this help menu.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-20s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
@@ -123,3 +124,7 @@ docker-build-hugo: hugo
 
 hugo:
 	git clone https://github.com/gohugoio/hugo.git --depth 1 -b v$(HUGO_VERSION)
+
+.PHONY: lychee-docker
+lychee-docker: gen-content
+	docker run --rm -it -e "GITHUB_TOKEN=$GITHUB_TOKEN" -v $$(pwd):/app $(LYCHEE_IMAGE_NAME) "/app/**/*.md"
--- a/lychee.toml
+++ b/lychee.toml
@@ -1,0 +1,46 @@
+# Verbose program output
+verbose = true
+
+# Show progress
+progress = false
+
+max_redirects = 10
+
+# Comma-separated list of accepted status codes for valid links.
+# Omit to accept all response types.
+accept = "200,204,302,301"
+
+
+# Exclude all private IPs from checking
+# Equivalent to setting `exclude_private`, `exclude_link_local`, and `exclude_loopback` to true
+exclude_all_private = true
+
+# Exclude all mail addresses from checking
+exclude_mail = true
+
+exclude = [
+    "(.*?).example.com",
+    "https://customdomain.com",
+    "https://hooks.slack.com/*",
+    "https://eventhubs.azure.net",
+    "https://proxy.corp",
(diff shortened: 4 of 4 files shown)


NissesSenap:
So markdown-link-check is much more restrictive than netlify is.
 
For example I did some tests with: content/en/docs/guides/image-update.md.
It gave me an error on `/use-cases/gh-actions-auto-pr`

It should be `../use-cases/gh-actions-auto-pr.md`, personally I agree with this, writing the correct path makes it much easier for me as a developer to know exactly where the file is, same thing to be forced to include file extension.

But It will be very annoying for lots of people in the beginning.

It's also possible to do cron jobs where we check all of the markdown files for broken changes: https://github.com/gaurav-nelson/github-action-markdown-link-check#scheduled-runs. 

But it will give us so many error in the beginning, personally I don't think it's worth to enable for now.
But if someone takes one for the team and brings up the files to the markdown-link-checks standard then it would be interesting to enable to make sure no links brakes thanks to time.

NissesSenap:
Another solution might be to use a plugin in netlify: https://github.com/munter/netlify-plugin-checklinks

dholbach:
I  had a look at the netlify plugin in #181 - it wasn't quite clear to me what exactly needed fixing, so I turned it off again as all the builds were ❌.

dholbach:
Can you make the output more verbose? I'd like to see which files and links the GH action actually checks.

Would there be a way to check the generated links in the HTML?

NissesSenap:
When you hit a error you will see it in the output, bellow you can find a example:
https://github.com/NissesSenap/website/runs/2806263094?check_suite_focus=true

I also updated the PR and set:
` use-quiet-mode: "no"`

Here you can find all config variabels:
https://github.com/gaurav-nelson/github-action-markdown-link-check#custom-variables

dholbach:
Not to be pedantic, but I can't quite see which files were checked in https://github.com/fluxcd/website/pull/320/checks?check_run_id=2819502707 (which was the last run).

NissesSenap:
I triggerd a few new PR:s but it dosen't look like `use-quiet-mode: "no"` makes any difference on the output :(
I see two ways forward, ether we trust that it checks the files that we can see in the PR since we are running: `check-modified-files-only` and think this is good enough.

In the error example PR we can see a good error:
image

We instead go over to running https://github.com/tcort/markdown-link-check from a pure github actions job with npm installed because there we can control the output much easier.

```
markdown-link-check content/en/docs/guides/image-update.md -v

FILE: content/en/docs/guides/image-update.md
[✓] https://kind.sigs.k8s.io/docs/user/quick-start/ → Status: 200
[✓] https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line → Status: 200
[✓] ../installation/ → Status: 200
[✓] https://github.com/stefanprodan/podinfo → Status: 200
[✓] mozilla-sops.md → Status: 200
[✓] sealed-secrets.md → Status: 200
[✓] ../components/image/imagepolicies.md#examples → Status: 200
[✓] ../use-cases/gh-actions-auto-pr.md → Status: 200
[✓] https://golang.org/pkg/text/template/ → Status: 200
[✓] webhook-receivers.md → Status: 200
[✓] ../components/notification/receiver.md → Status: 200
[✓] https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity → Status: 200
[✓] https://cloud.google.com/container-registry/docs/advanced-authentication#json-key → Status: 200
[✓] https://cloud.google.com/container-registry/docs/advanced-authentication → Status: 200
[✓] https://docs.microsoft.com/en-us/azure/aks/cluster-container-registry-integration → Status: 200
[✓] ../use-cases/azure.md#aad-pod-identity → Status: 200
[✓] https://github.com/fluxcd/flux2/tree/main/manifests/integrations/registry-credentials-sync/azure → Status: 200
[✓] https://docs.microsoft.com/en-us/azure/container-registry/container-registry-auth-kubernetes → Status: 200
[✓] https://docs.microsoft.com/en-us/azure/container-registry/container-registry-repository-scoped-permissions → Status: 200

19 links checked.
```
The bad thing here is that we have to write some script that notices changes that we changed in the PR + we have to maintain it.

Both works for me, I think that the github action is good enough but I can deferentially see why you want more output.

dholbach:
The output of the latter seems to be more helpful. Maybe we go with that one for now?

NissesSenap:
I haven't had time to test this PR in a good way. I will do that tomorrow with a few local PR:s in my own repo.

As the PR is setup now it will check for all files, including the ones that we get with `make gen-content`, so we might get errors that isn't even generated from this repo. I don't want us to not get any broken links errors when linking to external markdown files, I know I do this in the azure docs for example.

It should be noted that not only updated files is checked with the PR pipeline like it was in the previous example. This will most likely create errors in this check for people that update other parts of the homepage and the job can still fail.
It's possible to solve but will need to do some git magic together with forwarding those files in to the lychee part of the PR workflow.

Notice the GITHUB_TOKEN secret that is defined in the PR, it doesn't seem like you need to do anything with it. All we want to achieve is to not get rate limited towards github.com, for more information see: https://docs.github.com/en/actions/reference/authentication-in-a-workflow

I the cronjob I didn't enable any auto creation of issues like i have in the community repo. This due to that we will get allot of issues right now and I don't want it to generate a new set of issues every day.

We still only check in the markdwon files, we could also change over to checking the generated HTML, the good thing with checking the .md is that we will see issues that isn't included on the homepage and It will be easier to check where the actual error comes from.

I hope I have explained why I have maid the choices that I have made. I will give another update when I had time to do some more tests.

NissesSenap:
I have also started to take a closer look on the actual errors and I will need to include a config file where we exclude a number of endpoints.
So in short more works is needed, but if you already have feedback please share :)

NissesSenap:
So I have gotten the list of errors down by allot.

Even if I run with a GITHUB_TOKEN it seems like we git the request rate from github and get 429 errors.
I tried fixing this by lowering the number of concurrency requests `--max-concurrency=10` but it doesn't help, I guess the requests are still to fast.
I don't know if this works better if you have GitHub pro or if the requests comes from a GitHub action.

The funny thing is that it's doesn't say i hit any limits:
```shell
curl \
  -H "Accept: application/vnd.github.v3+json" \
  -u NissesSenap:$GITHUB_TOKEN https://api.github.com/rate_limit

{
  "resources": {
    "core": {
      "limit": 5000,
      "used": 794,
      "remaining": 4206,
      "reset": 1623870826
    },
    "search": {
      "limit": 30,
      "used": 0,
      "remaining": 30,
      "reset": 1623870578
    },
    "graphql": {
      "limit": 5000,
      "used": 0,
      "remaining": 5000,
      "reset": 1623874118
    },
    "integration_manifest": {
      "limit": 5000,
      "used": 0,
      "remaining": 5000,
      "reset": 1623874118
    },
    "source_import": {
      "limit": 100,
      "used": 0,
      "remaining": 100,
      "reset": 1623870578
    },
    "code_scanning_upload": {
      "limit": 500,
      "used": 0,
      "remaining": 500,
      "reset": 1623874118
    }
  },
  "rate": {
    "limit": 5000,
    "used": 794,
    "remaining": 4206,
    "reset": 1623870826
  }
}
```

Many on the adopters page seem to have some protection against curl, for example
```shell
curl https://www.offerup.com/
error code: 1020
```
A easy fix is of course to exclude that file but at at the same time it's one of those files where we can see companies diaper and it's kind of boring to have them around if they don't exist any more.


```txt
📝 Summary
---------------------
🔍 Total..........898
✅ Successful.....822
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded........65
🚫 Errors..........11

Errors in content/en/adopters.md
✗ https://www.offerup.com/ (HTTP status client error (403 Forbidden) for url (https://www.offerup.com/))
✗ https://yad2.co.il/ (HTTP status client error (403 Forbidden) for url (https://yad2.co.il/))
✗ https://www.loveholidays.com/ (HTTP status client error (403 Forbidden) for url (https://www.loveholidays.com/))
✗ https://apester.com/ (error sending request for url (https://apester.com/): error trying to connect: tcp connect error: Connection refused (os error 111))
✗ https://www.canva.com/ (HTTP status client error (403 Forbidden) for url (https://www.canva.com/))
✗ https://www.yad2.co.il/ (HTTP status client error (403 Forbidden) for url (https://www.yad2.co.il/))
✗ https://www.blablacar.com/ (HTTP status client error (403 Forbidden) for url (https://www.blablacar.com/))

Errors in content/en/docs/components/source/gitrepositories.md
✗ https://github.com/ (HTTP status client error (429 Too Many Requests) for url (https://github.com/))

Errors in content/en/security.md
✗ https://github.com/hiddeco (HTTP status client error (429 Too Many Requests) for url (https://github.com/hiddeco))

Errors in content/en/blog/2021-04-29-may-update.md
✗ https://hopin.com/events/gitops-con (error sending request for url (https://hopin.com/events/gitops-con): error trying to connect: tcp connect error: Connection refused (os error 111))

Errors in content/en/docs/faq/_index.md
✗ https://stefanprodan.github.io/podinfo2/index.yaml (HTTP status client error (404 Not Found) for url (https://stefanprodan.github.io/podinfo2/index.yaml))
````

**Comment by NissesSenap** [comment]
Many many tries later I finally manged to get only files changed in a PR to work, you can see the output here. https://github.com/NissesSenap/website/pull/2/checks?check_run_id=2856013440

**Comment by dholbach** [comment]
I see `grep: write error: Broken pipe` in the logs - not sure if this is an issue?

**Comment by NissesSenap** [comment]
You mean in the "Run make gen-content step"?

```
Run make gen-content
hack/generate-adopters-pages.py
hack/gen-content.py
hack/import-flux2-assets.sh

##########################                                                37.4%
######################################################################## 100.0%

######################################################################## 100.0%
grep: write error: Broken pipe
# Git Repositories
...
```
This is something that is from the hack/import-flux2-assets.sh script.
I had to do some google but my understanding is that we should be okay.
https://superuser.com/questions/554855/how-can-i-fix-a-broken-pipe-error

dholbach:
Thanks for chasing this up. Looks like we can land this now. We won't catch all the links, because we're not checking the generated HTML, but we check almost all. <3

dholbach (approved):

PR #329, opened 2021-06-15
Netlify fixes

hiddeco:
- Align Netlify Hugo version with Makefile
- Use available `URL` value during production-build


dholbach (approved):
thanks so much for figuring this out... maybe we can enable the link checking feature within netlify too...

stefanprodan (approved):

PR #338, opened 2021-06-18
Update source-watcher documentation

hiddeco:
- Bump prerequisites versions
- Show best-practice of using `utilruntime.Must` while registering
  API schemes


stefanprodan (approved):
I updated the source code in https://github.com/fluxcd/source-watcher/pull/21

hiddeco:
@stefanprodan there seems to be a difference between the Kustomize version advertised there and the bump I made here.

Reason I bumped it here is because of the breaking changes in Kustomize 4, assuming it being better to have people start of with the newest version so that they not have to deal with the changes.

PR #345, opened 2021-06-23
add Raspberry Pi FAQ - use a swap file

kingdonb:
Fix: fluxcd/image-reflector-controller#105

I'm not sure if there is anything we can do to make this better, if you got a truncated database log, you probably don't want to go on like everything is fine. It should land in CrashLoopBackOff, and you need to make sure there is enough memory for the operation to succeed. So, fixing it with a work-around and some documentation.


stefanprodan (approved):
LGTM

Thanks @kingdonb

Issue #324, opened 2021-06-14
"Other Flux projects" section items not centered


juozasg:
"GitOps Toolkit" and "Flagger" boxes are not centered. Especially noticeable on mobile.

hiddeco:
Noticed this too, another thing is that on ultra-wide screens (38" in my case), the page now goes outside the container making everything harder to read because I now have to move my head (or try to see sharp in the corners of my sight).

![image](https://user-images.githubusercontent.com/10063039/121950480-f6b87800-cd59-11eb-99ec-4f61bb90d689.png)

hiddeco:
A question related to the hero that has also changed: is the shadow on the H1 on purpose? If I disable it, it looks nicer (and less web 2.0?) to me.

juozasg:
Another, somewhat related issue is there is a narrow white vertical margin across the entire right side of the front page. It is not normally visible but scrolling the page to the right shows it. And it causes a horizontal scroll bar to appear on the bottom of the page. These alignment issues should be fixed in a same PR. 

![image](https://user-images.githubusercontent.com/4100/123028747-26403380-d39d-11eb-802d-998ec127fe8e.png)

juozasg:
@hiddeco I have copied your feedback on H1 drop shadow to #136.

PR #357, opened 2021-06-26
Front page item centering

juozasg:
Fixes ["Other Flux projects" section items not centered](https://github.com/fluxcd/website/issues/324)

Sets 1140px max-width for all front page sections to keep the layouts fixed on ultra-wide screens.


stefanprodan (approved):
LGTM 

Thanks @juozasg ☕

PR #358, opened 2021-06-28
Add documentation for how to use Ingress+Receivers

kingdonb:
In our docs for Webhook Receivers, we explain very briefly that you can use LoadBalancer or Ingress rules to get events from outside the cluster into the notification controller. The doc then goes on to show only a LoadBalancer example.

This PR expands the example to show both. There are other related changes I think we can make to this doc, perhaps hold off on merging this until we've had a chance to talk it out and ascertain how many of my expectations are off-base or if there's something important I missed.

### DIFF
--- a/content/en/docs/guides/webhook-receivers.md
+++ b/content/en/docs/guides/webhook-receivers.md
@@ -58,6 +58,30 @@ Wait for Kubernetes to assign a public address with:
 watch kubectl -n flux-system get svc/receiver
 ``` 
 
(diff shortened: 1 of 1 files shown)


kingdonb:
The other issues are mentioned in #359 to help keep this PR focused and mergeable (rather than boiling the ocean at once.)

stefanprodan (approved):
LGTM

Thanks @kingdonb ☕

Issue #188, opened 2021-04-30
Hard to find GH pages / repos


mewzherder:
AFAIK, I can't find it until I scroll to the bottom of the page and find the GH icon. Am I wrong?
It seems that visitors might expect a GH icon on the top bar as well, maybe to the left of "Docs"
Alternatively, perhaps it can be one of the options under the Project dropdown in the top menu.

PR #368, opened 2021-06-30
Add Github link

dholbach:
Also add logo to GH 

Fixes: #188

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -151,6 +151,13 @@ name = "Integrations"
 url = "/integrations"
 weight = 90
 
+[[menus.main]]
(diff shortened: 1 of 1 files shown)


dholbach:
https://deploy-preview-368--fluxcd.netlify.app/ - WDYT?

hiddeco:
Would add GitHub as the last menu item and the sole item with an icon as I find it now odd that the documentation item does have an icon but the others don't (probably because I am ready from LTR).

squaremo (approved):
I'm cool with this, thanks Daniel 🍎

Issue #369, opened 2021-06-30
Make buttons flat and crispy


dholbach:
Some feedback was that we should use flat buttons instead again. <https://flagger.app> was quoted as an example.

hiddeco:
![image](https://user-images.githubusercontent.com/10063039/123941575-93a22480-d99a-11eb-8041-3fc0923de240.png)

is a quick suggestion, but with the arrow icon added as Flagger has.

What I did to get there is change the background to `#fff`, drop all shadows around the text and button, and make the text color `#555`.

dholbach:
cc @juozasg

hiddeco:
This should probably be paired with a suggestion I made elsewhere to drop the shadow on the `h1` of the hero, which I can't find the issue / comment for at the moment.

dholbach:
#136

dholbach:
@hiddeco says
> There is still a 1px greenish border around the button that I have to remove / overwrite in my mock-up as well.

juozasg:
Removed green 1px border https://github.com/juozasg/website/commit/950accff21c7daf304e6bdcbca5f8e72b58a1cc1

dholbach:
I created a PR for you in #381.

Issue #208, opened 2021-05-05
adopters page: not all cards are equal height


dholbach:
![Screenshot from 2021-05-05 18-34-42](https://user-images.githubusercontent.com/1346979/117176821-989a8b80-add0-11eb-9ffc-7be7e0a69941.png)

In this case, the last entry being the only one in its `card-deck`.

juozasg:
Has this been fixed in another commit?

https://fluxcd.io/adopters/

dholbach:
Apparently. Well done whoever it was! :fireworks:

dholbach:
This is still an issue, from today:

![Screenshot from 2021-06-28 15-46-50](https://user-images.githubusercontent.com/1346979/123647139-1f477400-d828-11eb-881f-fdfa519499e4.png)

PR #376, opened 2021-07-01
adopters page card height and front page flat buttons

juozasg:
Fixes https://github.com/fluxcd/website/issues/208
Fixes https://github.com/fluxcd/website/issues/369

### DIFF
--- a/assets/scss/_adopters.scss
+++ b/assets/scss/_adopters.scss
@@ -21,6 +21,7 @@
     .card-body {
       align-items: center;
       display: flex;
+      min-height: 147px;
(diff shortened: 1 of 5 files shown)


dholbach (approved):
Nice work! :pretzel:

hiddeco:
There is still a 1px greenish border around the button that I have to remove / overwrite in my mock-up as well.

juozasg:
> There is still a 1px greenish border around the button that I have to remove / overwrite in my mock-up as well.

Fixed in juozasg:main

PR #389, opened 2021-07-06
Update installation.md

swapnild2111:
Corrected paths to point to correct patch yaml

### DIFF
--- a/content/en/docs/installation.md
+++ b/content/en/docs/installation.md
@@ -374,7 +374,7 @@ touch clusters/my-cluster/flux-system/gotk-components.yaml \
 ```
 
 Assuming you want to add custom annotations and labels to the Flux controllers,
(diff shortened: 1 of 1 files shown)


content/en/docs/installation.md
```diff
@@ -476,7 +476,7 @@ resources:
 - gotk-components.yaml
 - gotk-sync.yaml
 patches:
-  - path: psp-patch.yaml
+  - path: ../psp-patch.yaml
```
stefanprodan, line 479: The psp-patch file should in the same dir (`flux-system`) as this kustomization. Please update the path above to `clusters/my-cluster/flux-system/psp-patch.yaml` and undo this change.

stefanprodan:
@swapnild2111 please signoff your commit, see https://github.com/fluxcd/website/pull/389/checks?check_run_id=2996117649 for instructions. Thanks!

stefanprodan:
@swapnild2111 I ported your patch in #441 and added you as author. Thanks for the fix.

PR #371, opened 2021-06-30
disable foldable menu for now

dholbach:
Signed-off-by: Daniel Holbach <daniel@weave.works>

PR #390, opened 2021-07-06
Reenable foldable sidebar menu

dholbach:
Fixes: #371

A newer Docsy version will fix this and we can drop the override in `_nav.scss`.

### DIFF
--- a/assets/scss/blocks/_nav.scss
+++ b/assets/scss/blocks/_nav.scss
@@ -60,3 +60,10 @@ nav.navbar {
     font-weight: unset;
   }
 }
+
+// for now until https://github.com/google/docsy/issues/616 is fixed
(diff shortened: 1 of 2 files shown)


dholbach:
Screenshot from 2021-07-06 15-56-01

cc @stefanprodan

alisondy:
looks good however if you use keyboard navigation it will go into the nested items when they aren't visible

happy to approve and then look at this in the future

alisondy (approved):
looks good however if you use keyboard navigation it will go into the nested items when they aren't visible

approving however will need fix in future

dholbach:
> looks good however if you use keyboard navigation it will go into the nested items when they aren't visible

oh wow - sounds like like an upstream docsy issue...?

PR #394, opened 2021-07-07
fix other flux projects button aligment on mobile

juozasg:
Fix this:

image

Signed-off-by: Juozas <juozasgaigalas@gmail.com>


dholbach (approved):
well spotted and nicely done!

PR #404, opened 2021-07-12
De-emphasise "v2" in terminology somewhat

dholbach:
With moving the docs for f/flux and f/helm-operator
	under /legacy it should be even clearer now that the
	rest of the site is about the new version of flux:
	f/flux2.

	I left "v1" and "v2" in a couple of places where we
	refer to migration and do e.g. feature comparisons.

### DIFF
--- a/content/en/_index.html
+++ b/content/en/_index.html
@@ -3,7 +3,8 @@ title: Flux - the GitOps family of projects
 description: > 
   Flux is a set of continuous and progressive delivery solutions for Kubernetes, and they are open and extensible. 
   
-  The APIs of Flux are stable now. This means that now is a good time to start using the new version.
+  The APIs of Flux are stable now.
+  Now is the best time to start using Flux, or start your migration if you are a legacy user.
 ---
 
 {{% blocks/celebration emoji="🎉" url="/blog/2021/07/july-2021-update/#from-now-on-flux-apis-will-be-stable" %}}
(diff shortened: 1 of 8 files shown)


content/en/_index.html
```diff
@@ -3,9 +3,7 @@
 description: > 
   Flux is a set of continuous and progressive delivery solutions for Kubernetes, and they are open and extensible. 
   
-  Flux v2 will be GA within the next few months! 
-  
-  This means that Flux v1 will be deprecated before the end of 2021, so now is a good time to start using v2.
+  The APIs of Flux are stable now. This means that now is a good time to start using the new version.
```
alisondy, line 6: Rather than saying new version, start migrating from the legacy version.

```markdown
The APIs of Flux are stable now.
Now is the best time to start using Flux, or start your migration if you are a legacy user.
```

Something like that?

content/en/docs/_index.md
```diff
@@ -8,13 +8,13 @@ Flux is a tool for keeping Kubernetes clusters in sync with sources of
 configuration (like Git repositories), and automating updates to
 configuration when there is new code to deploy.
 
-Flux version 2 ("Flux v2") is built from the ground up to use Kubernetes'
+The new version of Flux is built from the ground up to use Kubernetes'
```
alisondy, line 11:
```suggestion
Flux is built from the ground up to use Kubernetes'
```

We shouldn't have to clarify new or old, in this context we are covering the current version of Flux.

alisondy (changes requested):
My thoughts
- Agree with keeping v1 v2 in the migration area
- Everywhere else where we have to reference Flux v1 we use legacy to describe
- only downfall with this is that it might be a tad confusing
have left some review notes/suggestionsw

alisondy (approved):

PR #409, opened 2021-07-13
[WIP] Add guide to setup GPG commit verification

fl42:
Hello,

This is my first documentation PR, feel free to give me feedback.

TODO:
- [ ] `flux bootstrap` commits are not signed, what is the best option to handle them? (`git commit --amend`?)


oliverbaehler:
Hi, just as little input how we are using gpg. 

The Kustomization for the `gitRepository` looks like this:

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
secretGenerator:
- name: catalog-gpg-pub-keys
  files:
  - ../../../gpg/member-1.asc
  - ../../../gpg/member-2.asc
  type: Opaque
generatorOptions:
 disableNameSuffixHash: true
``` 
And that's the `gitRepository` itself:

```
---
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: catalog
spec:
  interval: ${git_repo_interval:-${global_interval:-1m}}
 ....
  verify:
    mode: head
    secretRef:
      name: catalog-gpg-pub-keys
```

Maybe you could add this approach too.

kingdonb:
There is bootstrap signing

Thank you for contributing this, I noticed that the docs around commit signing and around `flux bootstrap git...` signing commits are a bit undercooked! Maybe we can still use some of this. It looks like commit signing in bootstrap was added some time after you contributed this PR, even if it was a while ago... https://github.com/fluxcd/flux2/pull/1854

I wrote some notes about my experience signing a commit with `flux bootstrap` here on another issue report:

https://github.com/fluxcd/flux2/issues/4519#issuecomment-2148159781

I think the keyring file is going to be the outlier, if we want the docs to be complete, it will be hard to cover all the different possibilities of whether you already have a keyfile or need to export one from the OS key store, but we can probably find an upstream document that covers all of that reasonably well (🤞)

PR #397, opened 2021-07-07
Landing Page Adopters Carousel [Draft]

juozasg:
Static draft preview of the adopters layout on the landing page. Carousel controls (Next/Previous buttons and Current Item Indicator) commented out to make it very simple and clean. Clicking the item links to the /adopters Page. 

What remains to be done:
* Clean up and factor out the markup
* Integrate with a YAML data source
* Modify/clone existing python scripts to generate YAML data weekly.  
* Update this to work with Landing Page section header color fixes that need to happen first

One idea: instead of generating 5 items weekly we could include all the items with logos and shuffle them randomly. 

@dholbach @hiddeco

PR #413, opened 2021-07-16
Add endorsements to main site

dholbach:
Closes: #397 

I had a bit of trouble keeping working across different branches, so I'm offering this PR instead of https://github.com/fluxcd/website/pull/397 now. I squashed a number of commits to make the log a bit more understandable.

List of changes:
- adopters yamls now specify if logo includes company name
- added endorsements to the site via its own yaml (right now 2 case studies with weave, 2 videos of talks)
- adopters script generates an endorsements carousel section with wall of logos in the background
- re-shuffled section on the site a little to have a more coherent flow


juozasg:
Static layout for the content (no carousel, no background) in my branch - at 3 different sizes:

image

image

image

juozasg:
Looks like this with the background, i don't like how the logos are cut in half by the text. I'll try to find a solution where background logos flow around the foreground text with responsive layout. I also tried to have text with transparent background but that will have a same problem with every logo with a white background.

image

juozasg:
Background disable for screen < 992px. Gradient on the sides only right now. Should be on the bottom too (or there should be nothing on the bottom). Going to integrate carousel animation before proceeding further to make sure it works well together. The blur might make transitions look distorted, although z-index tweaking might prevent. 

If this is not working well, we could also try a carousel with multiple logos (5 logos, or depending on screen width) horizontally that scrolls across all the available logos. 

image

image

dholbach:
This looks great.

juozasg:
Pushed to https://github.com/juozasg/website/tree/adopters-carousel-wip: a functioning carousel, tweaks to the background layout, a working gradient to fade the background, subtle opacity on the foreground and other adjustments.

This all works by keeping the carousel item external dimensions basically fixed at 600x600px. Next I will make sure this looks ok on mobile and fix Prev/Next slide button placement. The draft markup shouldn't change much from her eand can be generated now.

dholbach:
Thanks for your work on this! It looks great - I'll might plug in some of the other endorsements and see what things look like.

dholbach:
@juozasg Are you happy with this as well?

scottrigby (approved):
@dholbach @juozasg This looks fantastic ❤️ 🎉 👏  LGTM 🚀 

Looks like to add more endorsements, one just needs to make a pull request and add another list item to `content/en/endorsements.yaml` 👍

PR #416, opened 2021-07-19
FAQ: How to patch CoreDNS and other pre-installed addons?

HaveFun83:
I added a FAQ section which describe "How to patch CoreDNS and other pre-installed addons?"

### DIFF
--- a/content/en/docs/faq.md
+++ b/content/en/docs/faq.md
@@ -80,13 +80,13 @@ Depending on your configuration, a reconciliation can mean:
 * health checking of deployed workloads
 * garbage collection of resources removed from Git
 * issuing events about the reconciliation result
-* recoding metrics about the reconciliation process 
+* recoding metrics about the reconciliation process
 
 The 15 minutes reconciliation interval, is the interval at which you want to undo manual changes
 .e.g. `kubectl set image deployment/my-app` by reapplying the latest commit on the cluster.
 
 Note that a reconciliation will override all fields of a Kubernetes object, that diverge from Git.
-For example, you'll have to omit the `spec.replicas` field from your `Deployments` YAMLs if you 
+For example, you'll have to omit the `spec.replicas` field from your `Deployments` YAMLs if you
 are using a `HorizontalPodAutoscaler` that changes the replicas in-cluster.
 
 ### Can I use repositories with plain YAMLs?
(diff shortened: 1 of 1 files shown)


kingdonb:
Thanks for the PR! This is a good idea, I was doing something like this for myself the other day and didn't realize it wasn't documented. I have to get in the habit of going to look for a doc, rather than just shooting from the hip and guessing...

I was wondering if you ran into a problem with validation, that led you to write this part:

```
      containers:
      - name: coredns
```

Does this work how one would expect? I see it is valid Deployment YAML and I ran into this myself, noticing the `containers` key is required, so I defined the containers in full. I didn't consider that you can just include the containers by name and let the YAML merge that way, I assumed that if `containers` was present, the array contents would not be merged.

Like yourself as I suspect from reading, I only just wanted to add some annotations or ancillary parts (for me it was `nodeSelector` for my multi-arch Raspberry Pi cluster) so I really did not want to define the whole containers list and specify which images should be used, because that's like taking responsibility for keeping them up to date. (I was hoping my managed cluster runtime would take care of upgrading CoreDNS and friends for me when I do control plane upgrades. I don't want to do it myself, that's too much responsibility LOL... I actually didn't think that I even could do it like this.)

I'm guessing you tested and it worked as desired – rather than wiping out the existing `containers`, it just merges the array harmlessly by name, without overwriting or occluding any of the existing pod spec's container definition for the `coredns` container.

Do you think you can add some text or comment here on the PR about how it works, and if my assumptions are all valid?

HaveFun83:
> Thanks for the PR! This is a good idea, I was doing something like this for myself the other day and didn't realize it wasn't documented. I have to get in the habit of going to look for a doc, rather than just shooting from the hip and guessing...
> 
> I was wondering if you ran into a problem with validation, that led you to write this part:
> 
> ```
>       containers:
>       - name: coredns
> ```
> 
> Does this work how one would expect? I see it is valid Deployment YAML and I ran into this myself, noticing the `containers` key is required, so I defined the containers in full. I didn't consider that you can just include the containers by name and let the YAML merge that way, I assumed that if `containers` was present, the array contents would not be merged.
> 
> Like yourself as I suspect from reading, I only just wanted to add some annotations or ancillary parts (for me it was `nodeSelector` for my multi-arch Raspberry Pi cluster) so I really did not want to define the whole containers list and specify which images should be used, because that's like taking responsibility for keeping them up to date. (I was hoping my managed cluster runtime would take care of upgrading CoreDNS and friends for me when I do control plane upgrades. I don't want to do it myself, that's too much responsibility LOL... I actually didn't think that I even could do it like this.)
> 
> I'm guessing you tested and it worked as desired – rather than wiping out the existing `containers`, it just merges the array harmlessly by name, without overwriting or occluding any of the existing pod spec's container definition for the `coredns` container.
> 
> Do you think you can add some text or comment here on the PR about how it works, and if my assumptions are all valid?

Thanks a lot for your feedback.

I had a chat on slack with @stefanprodan about my problem before this PR. He gave me the hint of this option. I tired it in a fresh kubeadm bootstrapped cluster and it works as expected. As Flux will use kubectl 3-way-diff to make changes.

You can check the result via kubectl diff / kubectl apply before commit it into git.
The provided manifest is only an example and will work for kubeadm clusters.

stefanprodan (approved):
LGTM

Thanks @HaveFun83

PR #424, opened 2021-07-26
fix sortable image tag pattern for github run number example

Frederik-Baetens:
Signed-off-by: Frederik Baetens <baetens.fr@gmail.com>

The order of the sha & run number were swapped between the github actions workflow example and the pattern.


stefanprodan (approved):
LGTM

Thanks @Frederik-Baetens

PR #430, opened 2021-07-28
Image Update Automation is supported

kingdonb:
We have a couple of places where Image Update Automation is mentioned as "still currently unsupported, wait for the Feature Parity milestone" – I do not think this is true anymore, since feature parity has already been announced several minor releases ago.


stefanprodan (approved):
LGTM

Thanks @kingdonb

PR #414, opened 2021-07-19
Update helm.md

qingvincentyin:

PR #431, opened 2021-07-29
Fixed missing name on "flux create helmrelease"

or-shachar:
Without it the `flux create` would emit error:
```
✗ HelmRelease name is required
```


dholbach:
Thanks for your fix - #414 might be the more complete fix?

or-shachar:
Indeed. Closing in favor of #414

PR #445, opened 2021-08-06
changing the release-name/releaseName

vulndev:
As seen here https://github.com/bitnami-labs/sealed-secrets/issues/317#issuecomment-718608815 you should use the standard release name or you need to set an additional SEALED_SECRETS_CONTROLLER_NAME environment variable.


stefanprodan:
Thanks @vulndev can you please signoff your commit, see https://github.com/fluxcd/website/pull/445/checks?check_run_id=3261155999

vulndev:
> Thanks @vulndev can you please signoff your commit, see https://github.com/fluxcd/website/pull/445/checks?check_run_id=3261155999

Should be done now?

PR #448, opened 2021-08-06
added the standard controller name for kubeseal

vulndev:
As described here: https://github.com/bitnami-labs/sealed-secrets/issues/317#issuecomment-718608815 the standard controller name for kubeseal is sealed-secrets-controller and if you only use sealed-secrets you need to define another enrironment variable called: SEALED_SECRETS_CONTROLLER_NAME.

So as an alternative to my pull request you can also add the text for using this variable.


stefanprodan (approved):
LGTM 

Thanks @vulndev 🏅

Issue #449, opened 2021-08-09
Add events to landing page


dholbach:
https://fluxcd.io/community/#upcoming is too buried - let's find a way of highlighting key info on the landing page - maybe show upcoming meetings as well?

dholbach:
Maybe we could add upcoming talks/events to the Flux CNCF calendar too and use it as our source of truth?

@staceypotter @scottrigby @kingdonb WDYT?

PR #452, opened 2021-08-10
Read Flux meeting ical

dholbach:
Read the cncf-flux-dev calendar (ical) and show the events in the upcoming month.

Fixes: #449

Things that still need doing:

- [x] link to the calendar so folks know how to find out more (Daniel)
- [x] do timezone arithmetic so everything is in one TZ
- [x] make this look pretty (@juozasg)
- [x] decide as a group that we want to also keep track of talks, etc in our cncf-flux-dev calendar

---

~~- [ ] make sure everyone involved has access to the calendar~~ (moved to https://github.com/fluxcd/community/issues/113)
~~- [ ] hook up calendar include generation with daily cronjob~~ (should be handled by https://github.com/fluxcd/website/blob/main/.github/workflows/netlify.yml)
~~- [ ] figure out if we can link to the events in question... do we get the data in the ICS somewhere (Daniel)~~ - moved to #461
~~- [ ] make links clickable in the event description~~ - moved to #461
~~- [ ] have a mouse-over with event data (@juozasg + Daniel)~~ - moved to #461
~~- [ ] maybe convert to browser TZ(?)~~ - moved to #461

PR #456, opened 2021-08-12
Issue #449 calendar styling preview

juozasg:


kingdonb:
The specific issue that I've been presented with is that we need a link that we can use to get people to the latest information about our meetings, and it only gets us half of the way there to put it on the front page without a stable or permalink

https://deploy-preview-456--fluxcd.netlify.app/#td-block-10 does not seem to be a good way to link to this space on the webpage (1- it is not a good permalink, 2- it does not actually take you to the right place, or some scrolling happens after the page loads), and I am sure there is a better way. It would be ideal if we could link people to https://fluxcd.io/#calendar to get them directly to the calendar.

dholbach:
Closing this PR, all commits have been merged into #452

PR #467, opened 2021-08-17
Fix minor omission from Flux v1 docs

kingdonb:
This is the oldest open issue on the Flux v1 repo, and it can be closed with this docs update.

* fluxcd/flux#2998

Added one more:

* fluxcd/flux#3192

Signed-off-by: Kingdon Barrett <kingdon@weave.works>


stefanprodan (approved):
LGTM 

Thanks @kingdonb ☕

PR #472, opened 2021-08-20
Update doc about CoreDNS addon patching

nuriel77:
1. The example in the FAQ didn't work because the CoreDNS deployment is managed by EKS. EKS manages fields for affinity rules (see https://docs.aws.amazon.com/eks/latest/userguide/add-ons-configuration.html). This causes EKS to update the deployment once every 30 minutes (rolling upgrade) and flux will immediately revert that back, causing rolling upgrades to keep on happening on the cluster.

2. Another reason the example in the FAQ didn't work: this is an existing deployment deployed outside of Flux and we have to manage it differently as we must add the `selector` and `template` fields.

3. Changed the example to modify the `replicas` is this is a non-managed field and can be safely updated.

### DIFF
--- a/content/en/docs/faq.md
+++ b/content/en/docs/faq.md
@@ -175,8 +175,9 @@ the controller and validates the output using `kubeval`.
 ### How to patch CoreDNS and other pre-installed addons?
 
 To patch a pre-installed addon like CoreDNS with customized content.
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @nuriel77 🏅

Issue #462, opened 2021-08-13
roadmap: move outside the docs


dholbach:
@alisondy said: The Roadmap page might better live under https://fluxcd.io/roadmap - it's not necessarily a doc.

PR #475, opened 2021-08-23
Move roadmap out of docs

dholbach:
Fixes: #462


stefanprodan:
So looks like every time someone updates an internal link CI fails.

alisondy (approved):
LGTM

dholbach:
> So looks like every time someone updates an internal link CI fails.

No, it was an actual failure - I fixed it and rebased.

PR #487, opened 2021-08-29
Front page calendar details dropdowns

juozasg:
Instead of mousenter/mouseleave I'm using clicks on the calendar items to activate. This works with both touchscreen and mouse cursor devices. Please review the import-calendar.py changes. I'm doing some data cleaning up that maybe could be avoided by changing calendar data contents, but it should work and look neat with the existing data now.

### DIFF
--- a/hack/import-calendar.py
+++ b/hack/import-calendar.py
@@ -66,10 +66,21 @@ def write_events_html(events):
 
     for event in events:
         html += """
-        <tr>
+        <tr class="calendar-row">
             <td class="date">{}</td>
             <td class="time">{}</td>
             <td class="label">{}</td>
+        </tr>
+        <tr class="calendar-card">
+            <td colspan=3>
+                <dt>Where</dt>
+                <dd><a href="#">https://zoom.com/abcd</a></dd>
+
+                <dt>Organizer</dt>
+                <dd>Daniel Holback <a href="#">daniel@weave.works</a></dd>
+
+                <span class="description">Meetings minute agenda and videos <a href="#">https://docs.google.com/spreadsheets/d/1WcRgC7tXNpCK4AGm7fg37oesyVKr3THdJXQBlVq0rG8/edit#gid=220199767</a></span>
+            </td>
         </tr>
 """.format(
     event[0].strftime('%F'),
--- a/hack/import-calendar.py
+++ b/hack/import-calendar.py
@@ -62,33 +62,40 @@ def write_events_html(events):
         return
 
     html = """
-    <table>"""
+    <ul class="calendar-list">"""
 
     for event in events:
         html += """
-        <tr class="calendar-row">
-            <td class="date">{}</td>
-            <td class="time">{}</td>
-            <td class="label">{}</td>
-        </tr>
-        <tr class="calendar-card">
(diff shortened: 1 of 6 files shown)


dholbach:
Try changing the shebang to `python3` - maybe do that for all the scripts which don't have it yet.

I just used `netlify/build` to see what's available, it says:

```cli
buildbot@003a2a2e8c98:~/website$ python -V
Python 2.7.12
buildbot@003a2a2e8c98:~/website$
```

hack/import-calendar.py
```diff
@@ -40,20 +62,47 @@ def download_calendar():
...
 
+
+def read_organizer(event):
+    organizer = event['organizer']
+    email = organizer.title().split(':')[1].lower()
+    name = email
+    if 'cn' in organizer.params:
+        name = organizer.params['cn']
+
+    mailto_href = ''.format(email, name)
```
dholbach, line 73: Is this used anywhere?

juozasg: It is not. Wil delete. Ty!

hack/import-calendar.py
```diff
@@ -1,9 +1,10 @@
-#!/usr/bin/python
+#!/usr/bin/python3.7
```
dholbach, line 1:
```suggestion
#!/usr/bin/python3
```

This should be easier to fulfil on other setups, e.g. on Ubuntu 20.04:

```cli
❯ ls -l /usr/bin/python3
lrwxrwxrwx 1 root root 9 Mär 13  2020 /usr/bin/python3 -> python3.8
~ on ☁️  daniel@weave.works(europe-west1) 
❯ 
```

dholbach: This might fix the link checker action as well.

juozasg: I'm bit confused. I did this via a commit:

```
#!/usr/bin/python3
import sys
print(sys.version)
```
Commented out rest of the script. I got `3.5.7` from Netlify build log. Did you upgrade it elsewhere?

dholbach: This is strange...... so maybe indeed we need to find a way to specifically point to 3.7 somehow. I thought that's what `runtime.txt` would do: https://docs.netlify.com/configure-builds/manage-dependencies/#python

Maybe let's set `PYTHON_VERSION` in the `Makefile` to be sure?

dholbach: At some stage we should look into moving from Ubuntu 16.04 to Ubuntu 20.04 (just noticed that we're not on the newest yet). Will file a separate issue.

juozasg: I set `PYTHON_VERSION := 3.7` in Makefile. Still getting `3.5` from `print(sys.version)`

Makefile
```diff
@@ -23,6 +23,7 @@ HUGO_BIND_ADDRESS          ?= 127.0.0.1
 BUILDER_CLI                := docker
 # BUILDER_CLI                := okteto
 LYCHEE_IMAGE_NAME          ?= lycheeverse/lychee:202105190720247e4977
+PYTHON_VERSION						 := 3.7
```
dholbach, line 26: This can be dropped now.

PR #491, opened 2021-08-31
fix: branding link lead to 404

makkes:
Signed-off-by: Max Jonas Werner <mail@makk.es>

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -152,7 +152,7 @@ weight = 50
 [[menus.main]]
 name = "Branding"
 parent = "Project"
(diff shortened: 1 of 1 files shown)


dholbach:
Or maybe link to https://github.com/cncf/artwork/blob/master/examples/incubating.md#flux-logos - wdyt?

makkes:
> Or maybe link to https://github.com/cncf/artwork/blob/master/examples/incubating.md#flux-logos - wdyt?

Good idea! Thanks.

dholbach (approved):
Thanks a lot for the fix @makkes!

Issue #434, opened 2021-07-30
Declare less Hugo versions


dholbach:
We currently specify in many different places which version of hugo to use:

```cli
website on  main [$] via  v12.22.4 via 🐍 v3.8.10
❯ git grep 0.86.0
Makefile:HUGO_VERSION               ?= 0.86.0
docker-support/Dockerfile:FROM fluxcd/website:hugo-0.86.0-extended
netlify.toml:  HUGO_VERSION = "0.86.0"
package-lock.json:      "version": "0.86.0",
package-lock.json:      "resolved": "https://registry.npmjs.org/hugo-extended/-/hugo-extended-0.86.0.tgz",
package.json:    "hugo-extended": "^0.86.0",
website on  main [$] via  v12.22.4 via 🐍 v3.8.10
```

Right now, only `package*.json` are updated through dependabot. AFAIK, only `netlify.toml` is relevant for building the website? Can we slim this down somewhat?

@hiddeco @kingdonb Do you have any input on this?

kingdonb:
The convention in Ruby that I've seen to avoid duplicating the version is to put a file `.ruby_version` which is honored by some version managers, and use string interpolation to get that pulled into the Gemfile. That's not possible with package.json (and wouldn't work well for Dockerfile either.)

I think the best we can do is to build some CI job around the docker-support image, it should keep the published website support image updated on some cycle, and assert that versions in all of these places are correctly aligned. This can be something I could take a look at.

dholbach:
Is https://github.com/fluxcd/website/blob/main/Makefile#L126-L127 used at all anywhere?

yebyen:
Yes, this builds the base image since Hugo upstream doesn't provide one with the flags we need (or one per version at all)

PR #502, opened 2021-09-07
simplify hugo setup

dholbach:
- Use same pip call everywhere
-  Stop pulling in `hugo-extended` through `npm` - we don't use it.
	We already explain to install the extended version of hugo. `npm i` will install the hugo node module, but it doesn't
	install the binary in the PATH. For Netlify and the docker image we install hugo through other means.

	Overall this is a little improvement on #434, but likely not the best outcome yet.


dholbach:
After this is applied, we are down to these definitions of hugo versions:

```cli
❯ git grep "0\.88\.1"
Dockerfile:FROM fluxcd/website:hugo-0.88.1-extended
Makefile:HUGO_VERSION               ?= 0.88.1
netlify.toml:  HUGO_VERSION = "0.88.1"

website on  one-less-hugo-mention via  v14.17.6 via 🐍 v3.8.10 
❯ 
```

dholbach:
@kingdonb tested this and is happy to approve.

@stefanprodan Do you think you could merge and then run

```cli
rm -rf hugo/   # (to remove any stale artifacts from the previous image build)
make docker-push
```
to update the `fluxcd/website` images?

kingdonb:
I don't think it gets much better than this 👍 

I did:
```
rm -rf hugo/
make hugo
make docker-build-hugo
make docker-build-support
make docker-preview
```

and I am seeing the preview just fine on http://localhost:1313 👍 LGTM

@stefanprodan if you can update the image after this is done, then the docker-preview instructions should work again for anyone and match the latest version,

Once this merges to the main branch, if you check it out first, then this should be all you need to run, assuming you are also already logged in with credentials that can push to the `fluxcd/website` repo:

```
## Note: you may or may not have stale `hugo/` directory lying around already
rm -rf hugo/       # (to remove any stale artifacts from the previous image build)
make docker-push
```

This will build two images, and push both images. Here is the push target, for reference, so we can see when it has been updated:

https://hub.docker.com/r/fluxcd/website/tags?page=1&ordering=last_updated

People who run `make docker-preview` or `okteto up` will depend on this image, `fluxcd/website:hugo-support`

It is not versioned, and this is on purpose. It could probably be simplified somewhat. The build step takes quite a bit of time, and is still a manual step, but I think we don't update the base of the website that often, so hopefully that is OK. 👍

dholbach:
> I don't think it gets much better than this +1

@kingdonb How do you feel about something like this?

```diff
diff --git a/Makefile b/Makefile
index 3b4b11e..01c5590 100644
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ FONT_AWESOME_SEMVER_FOLDER := Font-Awesome-$(FONT_AWESOME_SEMVER)
 FONT_AWESOME_TARGET        := themes/$(DOCSY_COMMIT_FOLDER)/assets/vendor/$(FONT_AWESOME_SEMVER_FOLDER)
 
 DEV_IMAGE_REGISTRY_NAME    ?= fluxcd
-HUGO_VERSION               ?= 0.88.1
+HUGO_VERSION               ?= $(shell grep HUGO_VERSION netlify.toml | cut -d'"' -f2)
 HUGO_IMAGE_BASE_NAME       := website:hugo-$(HUGO_VERSION)-extended
 SUPPORT_IMAGE_BASE_NAME    := website:hugo-support
 HUGO_IMAGE_NAME            ?= $(DEV_IMAGE_REGISTRY_NAME)/$(HUGO_IMAGE_BASE_NAME)
```

kingdonb:
Yes, that change LGTM, unfortunately I'm fairly certain we can't perform the same or similar trick in the Dockerfile.

alisondy:
LGTM

alisondy (approved):

PR #509, opened 2021-09-09
Fix adopters carousel horizontal overflow scrollbar

juozasg:
100vw required for the carousel layout was sometimes producing a small empty margin on the right side of the page. This is an artifact of the browser calculation of viewport width when a vertical scrollbar is present. Hide horizontal scrollbar as a workaround.

### DIFF
--- a/assets/scss/_index.scss
+++ b/assets/scss/_index.scss
@@ -18,6 +18,11 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


dholbach (approved):
Thanks a lot @juozasg for fixing this!

PR #513, opened 2021-09-10
Review fonts 

juozasg:
Updated sizing and padding for fonts and surrounding elements to increase density of text in the front page so that it looks less zoomed out. Also made small changes to the H1 text that read better to me. Next step, is link colors/decorations to increase visibility. 

Please provide feedback

### DIFF
--- a/assets/scss/_index.scss
+++ b/assets/scss/_index.scss
@@ -34,22 +34,33 @@ main.td-main {
         padding-right: 0;
       }
     }
+    &.td-box {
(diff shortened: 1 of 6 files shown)


squaremo:
The text above the fold reads a little easier for me, I think that's the intended effect 👍

dholbach:
Bildschirmfoto von 2021-09-24 13-20-41

Not sure where this comes from, but I'm seeing this... ^

dholbach (approved):

PR #518, opened 2021-09-13
Break out bootstrap on installation page into separate guides

alisondy:
- Update installation guide to point to bootstrap section
- Update use cases to point to bootstrap page for Azure DevOps
- Break out bootstrap guides on installation page into seprate pages

Signed-off-by: Alison Dowdney <alison@alisondowdney.com>


stefanprodan:
Can we make the "Installation" into a section with bootstrap/upgrade/uninstall as sub-sections?

alisondy:
@stefanprodan 

currently the end tree of all these pr's is going to look like this
/content/en/docs/guides
         /install flux CLI
         /bootstrap flux
         /upgrade flux
         /uninstall flux

Bootstrap, Upgrade and Uninstall flux all belong together because they revolve around having the CLI already installed and interacting with, or deploying Flux on a Cluster

Install the CLI feels like it belongs on it's own as a primer to getting into those topics

Maybe something like ?
/content/en/docs/guides
         /install flux CLI
         /Deploy and Manage Flux
             /bootstrap flux
             /upgrade flux
             /uninstall flux

We can also look into the hierarchy after all the content is in too.

WDYT?

stefanprodan:
We moved the installation out of guides based on user feedback, I'm not keen on having all all these pages back to guides. Let's create a dedicated section please.

content/en/docs/bootstrap/airgapped.md
```diff
@@ -0,0 +1,116 @@
+---
+title: "Bootstrap Flux in an Air-gapped Environment"
```
dholbach, line 2: Can we use `title` and `linkTitle`? The titles in the sidebar get a little long?

dholbach: Potentially also use `weight`, Air-gapped is probably not the most searched for article?

content/en/docs/bootstrap/_index.md
```diff
@@ -0,0 +1,5 @@
+---
+title: "Bootstrap"
+linkTitle: "Bootstrap Flux"
+weight: 42
+---
```
dholbach, line 5: Do we want some text here to explain what's all done as part of the Bootstrap phase? And maybe link out to the `flux bootstap` CLI reference too, just in case?

content/en/docs/bootstrap/azuredo.md
```diff
@@ -0,0 +1,174 @@
+---
```
dholbach, line 1: Maybe rename to `azure-devops.md`? Might make it clearer from a URL perspective (SEO maybe too?).

content/en/docs/bootstrap/airgapped.md
stefanprodan: We should preserve the order from the original guide, Git/GitHub/GitLab/etc.

alisondy: Agree with linktitle and title, and weight

have avoided touching weight because there are alot of additions and i think we need to reweight a bunch of the pages

proposal - can we merge all 4 pr's into a seprate branch then do the reweighting in that branch then chuck it into main

content/en/docs/bootstrap/_index.md
alisondy: Yes but I think that should be done in a separate PR

content/en/docs/bootstrap/airgapped.md
alisondy: updated titles on all pages, - please review

Will address  page weights on separate PR

PR #524, opened 2021-09-13
Define hugo version once

dholbach:
Use ARG in Dockerfile to pass hugo version for image name.


stefanprodan (approved):

kingdonb:
I didn't think one could do that!

PR #528, opened 2021-09-16
Add new video to resources

staceypotter:
Added new video: Build a Smooth CI/CD Pipeline Using GitOps and Flux with Smaïne Kahlouch & Matheiu Gillot (at DevSecOps Road Trip: France) from 14 Sep 2021.

@dholbach - could you please verify I've done this correctly and it renders ok using the url & thumbnail instead of the youtube shortcut? I was trying to link it so that it would start at the correct timestamp. Thanks in advance! :) 
 
Signed-off-by: Stacey Potter <50154848+staceypotter@users.noreply.github.com>

### DIFF
--- a/content/en/resources.md
+++ b/content/en/resources.md
@@ -14,6 +14,12 @@ These resources are sorted by date - new entries are added to the top.
 {{% blocks/section color="white" %}}
 
 {{% blocks/resource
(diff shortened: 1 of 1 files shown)


dholbach (approved):
🎉

PR #550, opened 2021-09-27
Add tarioch/pre-commit-hooks

tarioch:


kingdonb:
Thank you for making this!

I don't think there's any reason we wouldn't accept this but I don't think it makes sense in the Integrations list. These are platforms built with flux. I opened up fluxcd/website#558 to make room for this kind of thing.

dholbach:
Merged through #741 - thanks @tarioch !

PR #564, opened 2021-10-01
Fix rewriting of links

dholbach:
Make sure to catch '(old-link)' and replace it with '(new-link)', so we don't expand the same link multiple times.


dholbach:
fixed in https://61572ded713853000828b0f1--fluxcd.netlify.app/community/#teams

hiddeco (approved):

PR #568, opened 2021-10-04
Separate the adopters logo wall from the carousel

juozasg:
Signed-off-by: Juozas <juozasgaigalas@gmail.com>

### DIFF
--- a/assets/scss/_adopters.scss
+++ b/assets/scss/_adopters.scss
@@ -18,37 +18,50 @@
    }
 }
 
-$gradient-width: 90px;
-$foreground-color: rgba($flux-white, 0.85);
+
+#adopters-wall {
+  @include media-breakpoint-down(md) {
+    display: none;
+  }
+
+  .adopters-label {
+    margin-top: 3rem;
+    margin-bottom: 1rem;
+
+    text-align: center;
+    font-size: 2rem;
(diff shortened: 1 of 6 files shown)


squaremo:
Yes, I like this -- we can have our logo wall _and_ our user testimonials :-D

dholbach:
yes, this looks good to me and makes things a little simpler!

what do you think about moving the "who is using flux" section further up?

hiddeco:
The logos seem to take up a lot of space on a MDPI screen (1280x800px, but screenshot is higher).

image

kingdonb:
I think that "wow look how many logos" is more or less the impression we're going for, this LGTM

I think the carousel works fine further down the page on its own, and the wall of logos is a lot more impactful as a stand-alone thing. Great work!

hiddeco:
> I think that "wow look how many logos" is more or less the impression we're going for

I do not agree, as there is more valuable information below the logo fold, and at some point the number of logos starts to look like a desperate upsell instead of an acknowledgement of trust.

hiddeco:
Or downscale them based on viewport (category)? I would keep the max height of the box of logos at 1/2 of the viewport at all times basically (no matter the number of rows). This also ensures the user knows the page continues.

PR #591, opened 2021-10-21
Fix FAQ example

kbreit:
The FAQ has an example about updating base Kubernetes manifests. I updated the example to add a top level `spec`.


kingdonb:
Thank you very much, 🥇 1st-time contributor gets a medal!

I think my review is sufficient but I cannot merge this, also I don't have an EKS cluster to test this on and verify it works from end to end. We uncovered this issue in a slack thread, here is a reference back to that thread:

https://cloud-native.slack.com/archives/CLAJ40HV3/p1634833189030200

PR #618, opened 2021-11-05
sops: Use minimal, decrypt-only policy for AWS

valorl:
The example IAM policy for AWS KMS integration is too wide for kustomize-controller which only really needs describe and decrypt. 

I've tested this by changing the policy in our dev environment, deleting a secret and seeing it be successfully applied the next reconcile.


stefanprodan (approved):
Thanks @valorl

PR #624, opened 2021-11-10
Add Virginia Tech to adopters

mikesir87:
Figured I'd add ourselves as an adopter now that we've given a talk or two about our use of it. Thanks!


dholbach (approved):
Thanks a lot for the PR @mikesir87 and welcome to the community!

PR #625, opened 2021-11-11
Add Audit blog post on fluxcd.io

scottrigby:
See https://www.cncf.io/blog/2021/11/11/flux-security-audit-has-concluded/
Should publish this immediately now, so all cncf blog post links work


stefanprodan (approved):
🚀

hiddeco:
```
6:03:03 PM: Error: Error building site: "/opt/build/repo/content/en/blog/2021-10-29-november-update.md:245:1": failed to render shortcode "tweet": failed to process shortcode: template: _internal/shortcodes/tweet.html:6:82: executing "_internal/shortcodes/tweet.html" at <index .Params 0>: error calling index: value has type int; should be string
```

squaremo (approved):
All looks present and correct :+1:

squaremo:
> error calling index: value has type int; should be string

There does seem to be an unrelated change in there; I thought it was harmless. Maybe not.

hiddeco (approved):

PR #627, opened 2021-11-16
Explain how to activate ImageAutomationPolicy

bigfleet:
As a Flux 1 user, we did not have these abstractions, and no comments were needed. 

This is better! 

I was aware of neither kyaml nor any other approach to comment templating on Kubernetes,  

This paragraph would have helped me, and I think it can help others.


yebyen:
This is a good change, IMHO - except that `ImageAutomationPolicy` is a new string that doesn't precisely refer to one thing, rather mixes a few concrete ideas. Either `ImageUpdateAutomation` or `ImagePolicy` are concrete things that are in the API, perhaps just saying "Image Automation Policy" without the quotes/pre-formatting would be an improvement?

bigfleet:
I will review verbiage to prevent confusion!

PR #637, opened 2021-11-19
Explain how to leverage image policy markers.

bigfleet:
As a Flux 1 user, we did not have these abstractions, and no comments were needed.

This is better!

I was aware of neither kyaml nor any other approach to comment templating on Kubernetes,

This paragraph would have helped me, and I think it can help others.

Signed-off-by: Jim Van Fleet <jim@jimvanfleet.com>


bigfleet:
Re-file of #627

stefanprodan (approved):
LGTM 

Thanks @bigfleet

Issue #598, opened 2021-10-26
Add security doc


dholbach:
#594 and #595 are relevant. We could also add information from https://fluxcd.io/docs/guides/repository-structure/ and around authentication.

dholbach:
#596 maybe too.

dholbach:
Refer to https://fluxcd.io/security as well please.

dholbach:
I am closing this issue now as all the individual tasks listed here have been resolved.

For new tasks, please use the `area/security` label: https://github.com/fluxcd/website/labels/area%2Fsecurity

Issue #595, opened 2021-10-25
Documenting the use of sensitive data by Flux in the cluster 


dholbach:
> Currently Flux provides no central documentation on the security of Flux. This is an issue as a user might expose more data than anticipated and currently has to read the implementation to identify these concepts. 

> Document the security context of Flux in a central location. Topics that are useful to document on such a page include: 
> 1. The sensitive data that Flux uses. 
> 2. The attack surface of Flux. 
> 3. What are the consequences of a compromise of Flux/component of Flux 4. Who should be allowed to administer Flux within an Organisation 
> 5. Security hardening of Flux, i.e. Flux can be used in many ways and what are the best practices from a security perspective and what are some of the pitfalls that can happen 

> Open source projects that have examples of this: 
> ● Docker: https://docs.docker.com/engine/security/ 
> ● Envoy: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/security
>    ○ In particular the threat model

scottrigby:
WIP for outlining how Flux correctly handles sensitive data during bootstrap, such as keys, and what the best practices are.

## Bootstrapping

- flux addresses 2 security:
- source (git repo, helm, buckets etc)
    - we have to talk about the security here, and what bootstrap does. 
    - asks for token for Github and GitLab etc
    - by default, we don't use token in cluster. it never leaves your machine
    - we use that to generate a deploy key, and we use that within the cluster
    - we're getting this question on slack a lot
    - but what if you do want flux to use a PAT? to give flux access to all repos in your org?
        - solution is you great a user for flux, and bootstrap using flux bot user token
        - then you can do this using the special flag in bootstrap `--token` and only if you do that, will flux use this in cluster
        - we must explain this in best practices
            - don't use your user PAT with --token
            - instead, use a Github machine user

scottrigby:
Issue closed by #641 and #711.

Will open a follow-up issue to address possible enhancements as mentioned in https://github.com/fluxcd/website/issues/595#issuecomment-976151045.

dholbach:
What from https://github.com/fluxcd/website/issues/595#issuecomment-976151045 is still missing?

scottrigby:
> What from [#595 (comment)](https://github.com/fluxcd/website/issues/595#issuecomment-976151045) is still missing?

Not necessarily missing. Elaborated here: https://github.com/fluxcd/website/issues/721#issuecomment-1021399424

PR #641, opened 2021-11-23
Initial PR for Flux security documentation landing page

### LINKED ISSUE #594: Documenting the permission system of Flux
> Flux and the GitOps toolkit leverages RBAC to control permissions. Flux creates a complex RBAC set up (and the set up based on the new impersonation will be even more complex) and this role/permission set up is not well-documented. In order to get an understanding of the RBAC set up it is currently required to inspect the implementation of RBAC yaml files themselves. We consider this to be a significant issue since the RBAC roles and Flux will be deployed within a cluster running potentially sensitive operations. 
 
> Provide a conceptual description of the permission system of Flux, preferably with schematic to make it easy for a reader to quickly understand what the roles, resources and permissions are in the Flux system. Highlight in particular: 
> - What RBAC roles are used 
> - What resources are used by the roles and the permissions on these resources
> - Whether Flux asks for more than necessary to ease implementation.
> - If the RBAC system is modifiable to create a more secure model. 
> - The effect that the RBAC implementation of Flux has on the entire state of the cluster.

### LINKED ISSUE #603: Explain RBAC roles situation
From Ada Logics

> The current controllers .yaml set up only offer cluster-wide roles (ClusterRole and ClusterRoleBinding). This is despite the controllers having the option of limiting monitoring to single namespaces through the use of the RUNTIME_NAMESPACE environment variable. In general, it seems that the roles in Flux are often default Kubebuilder roles and Kubebuilder overapproximates the permissions for the controllers. This is a known issue in Kubebuilder (see issue [here](https://github.com/kubernetes-sigs/kubebuilder/issues/1366)), in that Kubebuilder is opinionated and provides conservative permissions. However, this goes against the principle of least privilege.

> We want to emphasize here that we assume this is well-known by the Flux maintainers based on the description in the impersonation pull request (https://github.com/fluxcd/flux2/blob/1c5a25313561771d585c4192d7f330b45753cd99/docs/proposals/secure-impersonation.md#controller-serviceaccounts):
*“The controller ServiceAccounts are far overprivileged for Flux API operation”*

> Recommendation
> We recommend that Flux offers a solution where the controllers satisfy the principle of least privilege by providing a set of roles to be namespaces and only asks for verbs that each controller needs.

scottrigby:
- part of #598
    - note status this page is spelled out in an info alert, linking to this issue so users can follow progress to completion 
- fixes #594
- fixes #603
- added outline in progress to #595
    - the goal is to get out an initial version of this landing page, then add the additional sections in fast-follow PRs
- See markdown comments for editors


static/_redirects
```diff
@@ -9,3 +9,4 @@
 /legacy/helm-operator/guides/upgrade-to-ga    /legacy/helm-operator/how-to/upgrade-to-ga    301!
 /legacy/helm-operator/guides/upgrade-to-beta  /legacy/helm-operator/how-to/upgrade-to-beta  301!
 /docs/roadmap                                 /roadmap                                      301!
+/security                                     /docs/security                                301!
```
dholbach, line 12: I'm not sure I understand this.

My thinking was that we 
1. keep importing the security doc regarding processes into `/security`
2. add a new doc under `/docs/security` which explains Flux's security architecture and how to make Flux more security - which #598 is about

Both pages should link to each other.

content/en/docs/security/_index.md
```diff
@@ -0,0 +1,60 @@
...
+Please follow [this GitHub issue](https://github.com/fluxcd/website/issues/598) to stay updated on goals and progress as we complete this page.
+{{% /alert %}}
+
+## Introduction
+
+Flux has a multi-component design, and integrates with many other systems.
+
+This document outlines an overview of security considerations for Flux components, project processes, artifacts, as well as Flux configurable options and what they enable for both Kubernetes cluster and external system security.
+
+See our [security processes document](/security-processes) for vulnerability reporting, handling, and disclosure of information for the Flux project and community.
```
dholbach, line 20: Can we leave this at `/security`?

content/en/docs/security/_index.md
```diff
@@ -0,0 +1,60 @@
...
+Instead you specify a different service account in your manifest, and the Flux controllers will use the Kubernetes Impersonation API under `cluster-admin` to impersonate that service account.
+In this way, policy restrictions for this service account are applied to the manifests being reconciled.
+If the binding is not defined for the correct service account and namespace, it will fail.
+The roles and permissions for this multi-tenancy approach are described in detail here: https://github.com/fluxcd/flux2-multi-tenancy
+
+## External audit
+
+- In 2021 there was a security assessment and fuzzer development of the FluxCD repositories made by Ada Logics and funded by the OSTIF.
+- For a summary, please see [this blog post](https://fluxcd.io/blog/2021-11-10-flux-security-audit/)
+- For all details, please read the [audit full technical report](https://fluxcd.io/FluxFinalReport-v1.1.pdf)
```
dholbach, line 60: I wonder if this section shouldn't be under `/security` as well?

dholbach:
Thanks for looking into this - I like the text you've put together so far! I wonder how everyone else feels about this, but for me it'd make sense to have

- `/security` imported just like we do and have it explain 1) our project security processes, 2) CVEs that were discovered, 3) how to get in touch with the security team, 4) audits which were conducted and the like - so all in all organisational / project-wide / logistic bits 
- `/docs/security` as a landing page for everything security-related users of Flux need to know to make good decisions about how to integrate Flux securely into their setups and to understand which assumptions the Flux authors made

Does that make sense? If you agree, we could just keep the text you added and have it under `/docs/security` which would simplify the PR.

static/_redirects
scottrigby: Sure this sounds good, will revert a few commits

content/en/docs/security/_index.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Introduction
+
+Flux has a multi-component design, and integrates with many other systems.
+
+This document outlines an overview of security considerations for Flux components, project processes, artifacts, as well as Flux configurable options and what they enable for both Kubernetes cluster and external system security.
+
+See our [security processes document](/security-processes) for vulnerability reporting, handling, and disclosure of information for the Flux project and community.
+
+Note that while Flux is very active it is a vendor-neutral CNCF project maintained by volunteers, not by a single company. As such,  security issue handling is done on a best-effort basis.
```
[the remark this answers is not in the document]
scottrigby: good thinking

content/en/docs/security/_index.md
scottrigby: This is a good idea too 👍

scottrigby: I agree with the rest but I think this is important for end users who are assessing how secure Flux is. I think the other page is for security researchers to know how the actual process works for reporting, handling, and disclosures etc

content/en/docs/security/_index.md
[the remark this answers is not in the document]
scottrigby: Moved to https://github.com/fluxcd/.github/pull/11

content/en/docs/security/_index.md
dholbach: My thinking would be to have a very visible `note` at the top which recommends having at look at `/security` for  up-to-date info on CVEs/audits and other security project initiatives.

And then have `/docs/security` be a page which is  just documentation about Flux security.

That way we avoid having to update information in multiple places.

dholbach: WDYT?

config.toml
```diff
@@ -206,6 +206,12 @@ parent = "Project"
 url = "/roadmap"
 weight = 50
 
+[[menus.main]]
```
dholbach, line 209: This shouldn't be required - we already list the other Security doc under "Project".

external-sources/fluxcd/.github
```diff
@@ -1,2 +1,2 @@
-"/SECURITY.md","/security.md"
+"/SECURITY.md","/docs/security-processes.md"
```
dholbach, line 1: This can be dropped too.

content/en/docs/security/_index.md
scottrigby: Sure that works

config.toml
scottrigby: It wasn't there before, but believe this was just an oversight. Fixed the link 👍

scottrigby:
Hm, fixing DCO

scottrigby:
Alright, changes made. Once checks have run we should be good to go 👍

dholbach:
Looks great - can you squash?

content/en/docs/security/_index.md
```diff
@@ -0,0 +1,52 @@
+---
+title: "Security Documentation"
+linkTitle: "Security"
+description: "Flux Security documentation."
+---
+
+<!-- For doc writers: Step-by-step security instructions should live on the appropriate documentation pages. To fulfil our primise to end users, we should briefly outline the context here, and link to the more detailed instruction pages from each relevant part of this outline. -->
```
dholbach, line 7:
```suggestion
<!-- For doc writers: Step-by-step security instructions should live on the appropriate documentation pages. To fulfil our promise to end users, we should briefly outline the context here, and link to the more detailed instruction pages from each relevant part of this outline. -->
```

dholbach (approved):

content/en/docs/security/_index.md
scottrigby: Thx!

content/en/docs/security/_index.md
```diff
@@ -0,0 +1,52 @@
...
+
+Flux has a multi-component design, and integrates with many other systems.
+
+This document outlines an overview of security considerations for Flux components, project processes, artifacts, as well as Flux configurable options and what they enable for both Kubernetes cluster and external system security.
+
+See our [security processes document](/security) for vulnerability reporting, handling, and disclosure of information for the Flux project and community.
+
+## Controller permissions
+
+While Flux integrates with other systems it is built on Kubernetes core controller-runtime and properly adheres to Kubernetes security model including RBAC.
```
squaremo, line 24: Except for when it doesn't (cross-namespace references). That needs to be mentioned here.

content/en/docs/security/_index.md
```diff
@@ -0,0 +1,52 @@
...
+    - Can get, list, watch, create, patch, and delete configmaps and their status
+    - Can get, list, watch, create, patch, and delete coordination.k8s.io leases
+2. A `crd-controller` `ClusterRoleBinding`:
+    - References `crd-controller` `ClusterRole` above
+    - Bound to a service accounts for every Flux controller
+3. A `cluster-reconciler` `ClusterRoleBinding`:
+    - References `cluster-admin` `ClusterRole`
+    - Bound to service accounts for only `kustomize-controller` and `helm-controller`
+
+Flux uses these two `ClusterRoleBinding` strategies in order to follow the principle of least privilege, and to allow for clear access separation using tools purpose built for policy enforcement (OPA, Kyverno, admission controllers).
```
squaremo, line 44: The permissions needed by the controllers, and the permissions granted by the implementation as described above, are different. For example, the controllers don't _all_ need full access to _all_ of the API -- so it doesn't adhere to the principle of least privilege, for the sake of convenience.

dholbach:
@scottrigby ^ Do you think you could do a follow-up PR to this?

content/en/docs/security/_index.md
```diff
@@ -0,0 +1,52 @@
...
+
+For example, the design allows all controllers to access Flux CRDs (binds to `crd-controller` `ClusterRole`), but only binds the Flux reconciler controllers for Kustomize and Helm to `cluster-admin` `ClusterRole`, as these are the only two controllers that manage resources in the cluster.
+
+However in a soft multi-tenancy setup, Flux does not reconcile a tenant's repo under the `cluster-admin` role.
+Instead you specify a different service account in your manifest, and the Flux controllers will use the Kubernetes Impersonation API under `cluster-admin` to impersonate that service account.
```
squaremo, line 49: There is more detail to be given here: impersonation is only used for some client operations, and not for others. See the comment https://github.com/fluxcd/flux2/pull/2179/files#r763919841 for those things that are done with impersonation (basically, everything except for accessing sources).

squaremo:
Apologies for the late "review" -- I found my way here after attempting to document the same things in https://github.com/fluxcd/flux2/pull/2179. I shall consider following up on my comments with a PR ...

content/en/docs/security/_index.md
scottrigby: @squaremo Thanks, removing for now. 

However, do we have a list of these minimal permissions anywhere so that users who want to follow this principle will know the minimum privilege they require – even if that's not the default on bootstrap? I could see this being a potentially valuable option for some users. WDYT?

content/en/docs/security/_index.md
scottrigby: @squaremo I did not add the specific list here, since it seems to be Kustomize Controller specific, and your comment says you will put in a PR for that. If so, we will probably want to make a corresponding list of operations for Helm Controller too.

If you think we should spell out the exact operations here, let me know. Thinking that level of detail may be better in documentation in the the docs section for each controller, and we could link to them from here. WDYT?

PR #648, opened 2021-11-28
Search input text should be vertically centered

juozasg:
Replace this:
image

with:
image


dholbach (approved):
Nice one, thanks @juozasg !

PR #653, opened 2021-11-30
Add AWS CodeCommit use case page

Nalum:
Initial run at adding a setup of AWS CodeCommit with Flux.

I'm not sure if I've missed anything here, I need to run through it and make some notes for possible issues.

Any feedback at this point is appreciated.

### DIFF
--- a/content/en/docs/use-cases/aws-codecommit.md
+++ b/content/en/docs/use-cases/aws-codecommit.md
@@ -1,0 +1,184 @@
+---
+title: Using Flux on AWS With CodeCommit
+linkTitle: AWS CodeCommit
+description: "How to boostrap Flux on AWS EKS with CodeCommit Git repositories."
+weight: 10
+---
+
+## EKS Cluster Options
+
+### VPC CodeCommit Access
+
+If your VPC is setup without internet access or you would prefer that the access was over a private connection, you
+will need to set up a VPC endpoint to have access to CodeCommit. You can do this by following the guide [Using AWS
+CodeCommit with interface VPC endpoints](https://docs.aws.amazon.com/codecommit/latest/userguide/codecommit-and-interface-VPC.html).
+
+### Cluster Creation
+
+The following creates an EKS cluster with some minimal configuration that will work well with Flux:
+
+```sh
+eksctl create cluster
+```
+
+For more details on how to create an EKS cluster with `eksctl` please see [eksctl.io](https://eksctl.io).
+
+## Flux Installation for AWS CodeCommit
+
+The following replicates the [Flux bootstrap procedure](../installation.md#bootstrap) and represents
+the best practice for structuring the repository. For more information on the structure of the repository
+please see [Ways of structuring your repositories](../guides/repository-structure.md).
+
+Ensure you can login to [console.aws.amazon.com](https://console.aws.amazon.com) for your proper organization,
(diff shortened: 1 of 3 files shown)


content/en/docs/use-cases/aws-codecommit.md
```diff
@@ -0,0 +1,192 @@
...
+personal SSH keys, but take note that revoking the user's access to the repository will
+also revoke Flux's access. The better alternative is to create a machine-user whose sole purpose is
+to store credentials for automation. Using a machine-user also has the benefit of being able to be read-only or
+restricted to specific repositories if this is needed.
+
+{{% alert color="info" %}}
+Unlike `git`, Flux does not support the ["shorter" scp-like syntax for the SSH
+protocol](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_ssh_protocol)
+(e.g. `git-codecommit.<region>.amazonaws.com:v3`).
+Use the [RFC 3986 compatible syntax](https://tools.ietf.org/html/rfc3986#section-3) instead: `git-codecommit.<region>.amazonaws.com/v3`.
```
stefanprodan, line 104: Here is `amazonaws.com/v3` but above is `amazonaws.com/v1`, which one is it?

Nalum: It's `v1`. Will update it now :+1:

content/en/docs/use-cases/aws-codecommit.md
```diff
@@ -0,0 +1,192 @@
...
+## Cluster Creation
+
+The following creates an EKS cluster with some minimal configuration that will work well with Flux:
+
+```sh
+eksctl create cluster
+```
+
+{{% alert color="info" %}}
+When working with `eksctl` you can provide a `ClusterConfig`, using `-f`, to setup how you want your cluster to be
```
stefanprodan, line 17: Please replace this with something like "For more details on how to create EKS cluster with `eksctl` please see https://eksctl.io"

martinsmatthews:
This looks pretty similar to what I needed to do to get up and running - although I did also need to add a VPC endpoint for `git-codecommit` - thanks for documenting it!

As a flux newb it may be useful to explain the dir structure used in this example - unless that is a standard flux thing and this is explained elsewhere. Calling several things `flux-system` is also somewhat confusing but again that may be a standard thing?

Finally, it would be nice to have some pointers on where to go next as this just gets the flux "infrastructure" components into the repo but as an end user I'm interested in how I then get my HelmReleases etc in there as well - should I use another kustomization or will the existing one pick up any new kustomization.yaml for the cluster. Also then adding new clusters, assuming I'm operating more than one. Maybe just some links to further reading/next steps.

stefanprodan:
@martinsmatthews good point, the doc should state that this replicates the structure from `flux bootstrap` and represents the best practice around organising repos. This doc should like to:
- Flux bootstrap procedure https://fluxcd.io/docs/installation/#bootstrap 
- Ways of structuring your repositories https://fluxcd.io/docs/guides/repository-structure/

Nalum:
Thanks @martinsmatthews and @stefanprodan I'll look at updating the doc with this information :+1:

stefanprodan (approved):
LGTM

Thanks @Nalum can you please squash the commits and rebase with upstream main please.

PR #679, opened 2021-12-09
Add permissions info for GitHub installation

SamLR:
Add some details on the permissions needed to bootstrap pre-existing repositiories.

This is instead of the change suggested here https://github.com/fluxcd/flux2/pull/2197


stefanprodan (approved):
LGTM

Thanks @SamLR

SamLR:
> LGTM
> 
> Thanks @SamLR

Thanks @stefanprodan  : )

PR #687, opened 2021-12-15
Update theme (and its dependencies)

dholbach:


dholbach:
An update of `docsy` and its dependencies. Applying upstream changes to over our partial overrides.

kingdonb (approved):
LGTM 👍

kingdonb:
The updated support images have been pushed, so `make docker-preview` and its friends will go on working after this. 👍 

Any docs changes that are prior to this PR should be rebased in order to test effectively against the newer versions.

PR #698, opened 2022-01-07
Fix: Added space to correct code parsing

santschi:
/" is interpreted as an escaped character, so the rest was shown as part of the string.

### DIFF
--- a/content/en/docs/use-cases/jenkins.md
+++ b/content/en/docs/use-cases/jenkins.md
@@ -136,7 +136,7 @@ pipeline {
             developmentTag = "${branchName}-${gitCommit}-${unixTime}"
             developmentImage = "${dockerRepoUser}/${dockerRepoProj}:${developmentTag}"
           }
-          sh "docker build -t ${developmentImage} ./"
+          sh "docker build -t ${developmentImage} ./ "
         }
       }
     }


kingdonb:
Where do you see this rendering error? It doesn't look like anything is wrong on https://fluxcd.io/docs/use-cases/jenkins/#example-jenkinsfile to me:

Screen Shot 2022-01-07 at 10 47 55 AM

santschi:
Yes, everything up to this exact point is correct.
But if you look at any point afterwards it is colored as part of a string. In effect everything after 
`./"` is still considered part of the` sh " .... `string

kingdonb:
I see what you're saying now, the syntax highlighter is having a hard time it looks like because it's somehow not configured for Groovy, I think, or my Jenkinsfile doesn't evaluate through the syntax highlighter correctly as Groovy code, perhaps.

I remember looking at this issue when I merged it, I think what's going on is the syntax highlighter has a bug and doesn't understand slash as a binary operator, it looks like every instance of a slash is interpreted as a slashy string. That's not how Groovy interprets the file, so I think this is a bug in the syntax highlighter. Not sure what else we can do about it.

The change doesn't appear to fix it either way, if you have a look at the deploy-preview! Also note, we cannot merge PRs without the DCO sign-off. Thank you for contributing! I'm not even sure exactly where to send this issue, if hugo farms out syntax highlighting to another module, or where the fix might belong, but this is unfortunately not the correct fix.

kingdonb:
I did some research, now at least I know where the bug report would go.

There are two code highlighting engines in Hugo, `goldmark` and `PrismJS`

I found that PrismJS does support highlighting Groovy really well, while goldmark does not, at least not perfectly wrt slashes and slashy strings.

Here's the PrismJS look:

Screen Shot 2022-01-07 at 2 52 22 PM

The bad news is, while Docsy has PrismJS support, it only bundles a minimal set of languages in the base theme and Groovy is not one of them. So even if we decided to use PrismJS, which is provided with support from Docsy our chosen theme, we're on the hook for maintaining a copy of PrismJS with any additional languages we choose to add, like Groovy here.

That's a bigger change than I'm prepared to make or propose, to solve this. I'm inclined to punt and cross our fingers that Goldmark improves their highlighting for groovy. It looks like they don't mention Groovy at all, so I'm inclined to think it isn't really using a Groovy parser or highlighter (and they might not even consider this a bug.)

We could just remove the highlighting altogether, if it's distracting that it is highlighted wrong? WDYT

santschi:
I would definitely not remove the syntax highlighting. Even if it's not perfect it delineates the code very well from other content.

As far as I know goldmark is based on [chroma](https://github.com/alecthomas/chroma) which supports Groovy.

But honestly I do not think this issue is worth sinking hours of your time into. I just thought I had a quick fix and was wrong.

kingdonb:
No worries, you are right. I just found that too: chroma is the highlighter, and it does have Groovy support, I'll file the bug there :)

I needed to learn more about the website generator / docsy engine anyway, you're not putting me out at all!

kingdonb:
The repro steps are filed upstream at:

* alecthomas/chroma#588

Thanks for using Flux 🥇

kingdonb:
Turns out there was an easy fix after all. It's not what I thought it was.

The `chroma` highlighter's `groovylexer` does not appear to consider a `/` front slash as a symbol that can be used as a binary operator, so any slashes that occur outside of a string will be considered slashy strings (they can be multi-line too!)

What a nightmare to parse. For the Flux docs, I will resolve this issue permanently by avoiding using the `/` as a divide op.

🔥

PR #699, opened 2022-01-07
Based on #698

kingdonb:
There is something wrong with the syntax highlighter here, testing a theory to riff off of #698 which attempted to fix it, but didn't seem to work.


kingdonb:
Neither of us have guessed correctly about what causes this issue.

I now believe that the syntax highlighting issue is related to slashy strings. Parsing Groovy must be very difficult. 🤷

PR #700, opened 2022-01-07
Remove double division and cast

kingdonb:
I am not much of a Groovy programmer, so when I wrote this I did not realize there was another way to do integer division.

This is just dense enough groovy syntax and hard to parse for chroma to balk at it, this resolves the issue from #698 in lieu of an upstream fix for alecthomas/chroma#588


kingdonb:
Better!
Screen Shot 2022-01-07 at 4 24 48 PM

Compare to: https://github.com/fluxcd/website/pull/698#issuecomment-1007513758

dholbach (approved):
Not a groovy person, but looks legit...!

PR #702, opened 2022-01-11
Add docs for autologin(ACR/ECR/GCR)

SomtochiAma:
This pull request adds documentation for GCR/ACR auto-login Integration and also moves the docs on using cron jobs to a separate page.

Part of: https://github.com/fluxcd/flux2/issues/2308

Signed-off-by: Somtochi Onyekwere <somtochionyekwere@gmail.com>

### DIFF
--- a/content/en/docs/guides/cron-job-image-auth.md
+++ b/content/en/docs/guides/cron-job-image-auth.md
@@ -103,7 +103,7 @@ spec:
             - -ce
             - aws ecr get-login-password --region ${REGION} > /token/ecr-token
           containers:
-          - image: bitnami/kubectl
+          - image: ghcr.io/fluxcd/flux-cli:v0.25.2
             name: create-secret
             imagePullPolicy: IfNotPresent
             env:
--- a/content/en/docs/guides/cron-job-image-auth.md
+++ b/content/en/docs/guides/cron-job-image-auth.md
@@ -1,0 +1,377 @@
+---
+title: "How to use cron jobs to sync image repository credentials"
+linkTitle: "cron jobs for image repository cretentions"
+description: "How to use cron jobs to sync image repository credentials."
+weight: 90
+---
+
+## Image Repository Authentication
+
+While [native authentication mechanisms](image-update.md/#imageRepository-cloud-providers-authentication) 
+are available, using a cron job is the preferred way of syncing image repository credentials for
+multi-tenancy as the controller cannot natively get access to the image repository.
+
+### AWS Elastic Container Registry
+
+#### Using CronJob to sync ECR credentials as a Kubernetes secret
+
+The registry authentication credentials for ECR expire every 12 hours.
+Considering this limitation, one needs to ensure the credentials are being
+refreshed before expiration so that the controller can rely on them for
+authentication.
+
+The solution proposed is to create a cronjob that runs every 6 hours which would
+re-create the `docker-registry` secret using a new token.
+
+Edit and save the following snippet to a file
+`./clusters/my-cluster/ecr-sync.yaml`, commit and push it to git.
+
+```yaml
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: ecr-credentials-sync
+  namespace: flux-system
+rules:
+- apiGroups: [""]
+  resources:
+  - secrets
+  verbs:
+  - get
+  - create
+  - patch
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: ecr-credentials-sync
+  namespace: flux-system
+subjects:
+- kind: ServiceAccount
+  name: ecr-credentials-sync
+roleRef:
+  kind: Role
+  name: ecr-credentials-sync
+  apiGroup: ""
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ecr-credentials-sync
+  namespace: flux-system
+  # Uncomment and edit if using IRSA
+  # annotations:
+  #   eks.amazonaws.com/role-arn: <role arn>
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: ecr-credentials-sync
+  namespace: flux-system
+spec:
+  suspend: false
+  schedule: 0 */6 * * *
+  failedJobsHistoryLimit: 1
+  successfulJobsHistoryLimit: 1
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: ecr-credentials-sync
+          restartPolicy: Never
+          volumes:
+          - name: token
+            emptyDir:
+              medium: Memory
+          initContainers:
+          - image: amazon/aws-cli
+            name: get-token
+            imagePullPolicy: IfNotPresent
+            # You will need to set the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables if not using
+            # IRSA. It is recommended to store the values in a Secret and load them in the container using envFrom.
+            # envFrom:
+            # - secretRef:
+            #     name: aws-credentials
+            env:
+            - name: REGION
+              value: us-east-1 # change this if ECR repo is in a different region
+            volumeMounts:
+            - mountPath: /token
+              name: token
+            command:
+            - /bin/sh
+            - -ce
+            - aws ecr get-login-password --region ${REGION} > /token/ecr-token
+          containers:
+          - image: bitnami/kubectl
+            name: create-secret
+            imagePullPolicy: IfNotPresent
+            env:
+            - name: SECRET_NAME
+              value: ecr-credentials
+            - name: ECR_REGISTRY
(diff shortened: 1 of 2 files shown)


kingdonb:
I'm testing the Google Cloud Workload Identity work, I ran into this error applying one of the patches from these examples:

```
02-gke-cluster-1      	False	kustomize build failed: add operation does not apply: doc is missing path: "/metadata/annotations/spec/": missing value	                                                   	False
```

I suspect it would work fine if I had used patchesStrategicMerge instead: https://github.com/kubernetes-sigs/kustomize/issues/2986#issuecomment-847125775

tl;dr from the link above, json patch 6902 cannot really be used to add a path that does not exist, if you have a hash then you can add a field to it, but if there is no hash "annotations" which I think there generally isn't, it will not be created in a json patch. Maybe this works if you annotate the service account by hand with the `gcloud` CLI, I'm still new at this.

I tried this patch instead and it worked:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- gotk-components.yaml
- gotk-sync.yaml
patches:
- target:
    version: v1
    group: apps
    kind: Deployment
    name: image-reflector-controller
    namespace: flux-system
  patch: |-
    - op: add
      path: /spec/template/spec/containers/0/args/-
      value: --gcp-autologin-for-gcr
patchesStrategicMerge:
- |-
  apiVersion: v1
  kind: ServiceAccount
  metadata:
    name: image-reflector-controller
    namespace: flux-system
    annotations:
      iam.gke.io/gcp-service-account: kingdon-test-gcr-sa@dx-kingdon.iam.gserviceaccount.com
images: # []
- name: ghcr.io/fluxcd/image-reflector-controller
  newName: docker.io/somma/image-reflector-controller
  newTag: test-autologin-18df01042
```

I'm sending you some corrections for the issues I ran into 👍 the good news is, I was able to confirm the GKE workload identity can access my `gcr.io/dx-kingdon/podinfo` private repository and sync tags without issue:

```
$ flux get image repository
NAME   	READY	MESSAGE                      	LAST SCAN                	SUSPENDED
podinfo	True 	successful scan, found 2 tags	2022-01-14T09:02:47-05:00	False
```

I know I'm annotating the SA correctly now, because when it was incorrect I got a 400 "bad request" error. There is a `.` instead of an `@` in your example annotation. I found references to all of the correct information in the guide you linked here (the example you gave there is also looks to be more correct):

* https://github.com/fluxcd/image-reflector-controller/pull/194#issuecomment-1009777498

content/en/docs/guides/image-update.md
```diff
@@ -627,396 +627,99 @@ flux create image policy podinfo \
...
+- target:
+    version: v1
+    group: ""
+    kind: ServiceAccount
+    name: image-reflector-controller
+    namespace: flux-system
+  patch: |-
+    - op: add
+      path: /metadata/annotations/spec/
+      value: <gcp-service-account-name>.<PROJECT_ID>.iam.gserviceaccount.com
```
kingdonb, line 702:
```suggestion
- |-
  apiVersion: v1
  kind: ServiceAccount
  metadata:
    name: image-reflector-controller
    namespace: flux-system
    annotations:
      iam.gke.io/gcp-service-account: <gcp-service-account-name>@<PROJECT_ID>.iam.gserviceaccount.com
```

content/en/docs/guides/image-update.md
```diff
@@ -627,396 +627,99 @@ flux create image policy podinfo \
...
+    - op: add
+      path: /spec/template/spec/containers/0/args/-
+      value: --gcp-autologin-for-gcr
+ ### add this patch to annotate service account if you are using Workload identity
+- target:
+    version: v1
+    group: ""
+    kind: ServiceAccount
+    name: image-reflector-controller
+    namespace: flux-system
```
kingdonb, line 698:
```suggestion
patchesStrategicMerge:
```

I could only make this patch work as a `patchesStrategicMerge`

content/en/docs/guides/image-update.md
kingdonb: Note the `@` replacement

I tested several incorrect formations of the `patchesStrategicMerge` before landing on this one 🙄 please double check this is correct, I believe I have it now.

content/en/docs/guides/cron-job-image-auth.md
```diff
@@ -0,0 +1,377 @@
...
+
+#### Using Native GCP GCR Auto-Login
+
+#### Using access token [short-lived]
+
+{{% alert color="info" title="Workload Identity" %}}
+Please ensure that you enable workload identity for your cluster, create a GCP service account that has
+access to the container registry and create an IAM policy binding between the GCP service account and
+the Kubernetes service account so that the pods created by the cronjob can access GCP APIs and get the token.
+Take a look at [this guide](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity)
```
kingdonb, line 163: As a newbie to GCP Workload Identity, I was able to follow the abridged GCR instructions solely because of this link, it was too complicated without all four main steps being laid out explicitly as done here. I also could have used a primer on GCP IAM roles, as I'm pretty sure it's possible to get this part wrong, but the wizard seems to have created the right thing for me based on selecting "Container Registry Service Agent".

Maybe we should not be in the business of onboarding GKE customers to their IDP, but after this feature merges I'd like to take a stab at a guide that will cover this information in more detail again.

In the mean time, we should make sure the "[this guide](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity)" link is included in the Workload Identity / Autologin instructions below as well. IMHO 👍 this is a good structure for this change, and it makes it clear we understand that autologin is likely to be the preferred and ideal mode for most users.

Even if it is not ideal for (soft) multi-tenancy, I think that is the reality on the ground.

content/en/docs/guides/image-update.md
```diff
@@ -627,396 +627,99 @@ flux create image policy podinfo \
...
+```
+
+#### Using Native GCP GCR Auto-Login
+
+There is native support for the GCP Google Container Registry available since `image-reflector-controller` [vadd-version](link to changelog)
+which was released with Flux release v[add-flux-ver]. This feature is enabled by setting the `--gcp-autologin-for-gcr`
+flag. This works with both clusters that have Workload Identity enabled, and those that use the default service account.
+This flag can be added by including a patch in the `kustomization.yaml` overlay file in your `flux-system`,
+similar to the process described in [customize Flux manifests](../installation.md/#customize-flux-manifests):
+
```
kingdonb, line 679: This would be a good place to mention/add the "this guide" link I mentioned before.

kingdonb (changes requested):
👍 please adopt the suggestions or LMK if there are other changes

I am really pleased that I was able to make this work, it was not easy, but the information in this comment was super concise and very helpful to my overall understanding here:

* https://github.com/fluxcd/image-reflector-controller/pull/194#issuecomment-1009777498

kingdonb:
I can adopt this branch myself, and reissue it with my suggestions incorporated as a new PR (if that helps!)

stefanprodan:
@SomtochiAma please add @kingdonb to your fork so he can push to to your branch.

SomtochiAma:
@kingdonb I have sent out an invite to my fork of this repo

content/en/docs/guides/cron-job-image-auth.md
kingdonb: Resolved with 1f9240b

kingdonb:
My approval is pending the merge of the following PR:

* https://github.com/fluxcd/image-reflector-controller/pull/194

The draft status of this PR reflects that the depended on PR has not been merged yet.

content/en/docs/guides/cron-job-image-auth.md
```diff
@@ -0,0 +1,377 @@
...
+              value: us-east-1 # change this if ECR repo is in a different region
+            volumeMounts:
+            - mountPath: /token
+              name: token
+            command:
+            - /bin/sh
+            - -ce
+            - aws ecr get-login-password --region ${REGION} > /token/ecr-token
+          containers:
+          - image: bitnami/kubectl
```
stefanprodan, line 106:
```suggestion
          - image: ghcr.io/fluxcd/flux-cli:v0.25.2
```

content/en/docs/guides/cron-job-image-auth.md
```diff
@@ -0,0 +1,377 @@
+---
+title: "How to use cron jobs to sync image repository credentials"
+linkTitle: "cron jobs for image repository cretentions"
```
stefanprodan, line 3:
```suggestion
linkTitle: "Configure image automation authentication"
```

content/en/docs/guides/cron-job-image-auth.md
```diff
@@ -0,0 +1,377 @@
+---
+title: "How to use cron jobs to sync image repository credentials"
```
stefanprodan, line 2:
```suggestion
title: "Configure image automation authentication"
```

stefanprodan (changes requested):

stefanprodan (approved):
LGTM

I'll make the title changes in another PR.

Issue #596, opened 2021-10-25
Release cycle information and expectation


dholbach:
We have https://github.com/fluxcd/flux2/blob/main/docs/internal/release.md but we currently are lacking user-targetted

> security-relevant information such as how long versions are supported and how often to expect releases. 
> Examples of open source projects that describe their release cycle: 
> ● Kubernetes: https://kubernetes.io/releases/version-skew-policy/ 
> ● Redis: https://redis.io/topics/releases 
> ● Linux kernel: https://www.kernel.org/category/releases.html

Maybe make this a section in a top-level `/docs/security` page?

pjbgf:
We already have some of this information in the [Migration Timetable page](https://fluxcd.io/docs/migration/timetable/) (e.g. support / version skew policy). 

But thing such as release cycle, version support and skew policy deserve a standalone page. 
This make even more sense for folks new to flux, outside the "migration path".

PR #716, opened 2022-01-25
Add FAQ for when to expect release schedule info

priyanka-ravi:
resolves fluxcd/website#596

### DIFF
--- a/content/en/docs/faq.md
+++ b/content/en/docs/faq.md
@@ -390,4 +390,4 @@ to help you migrate to Flux v2:
 
 ### Where can I find information on how long versions are supported and how often to expect releases?
 
-Releases are as needed for now and a release page with the schedule will be created when Flux reaches general availability status ([Flux roadmap can be found here](https://fluxcd.io/roadmap/)). We understand the importance of having a release schedule and are aware that such a schedule will support the users' ability to plan for an upgrade in advance.
+Releases are as needed for now and a release page with the schedule will be created when Flux reaches general availability status ([Flux roadmap can be found here](https://fluxcd.io/roadmap/)). We understand the importance of having a release schedule and are aware that such a schedule will support the users' ability to plan for an upgrade in advance. For now you can expect the current release cadence of at least once per month to continue.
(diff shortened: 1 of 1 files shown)


dholbach:
Can you please sign your commits like so...?

In your local branch, run: git rebase HEAD~1 --signoff
Force push your changes to overwrite the branch: git push --force-with-lease origin release-schedule-faq

content/en/docs/faq.md
```diff
@@ -385,3 +385,9 @@ to help you migrate to Flux v2:
 - [How to manage multi-tenant clusters with Flux v2](https://github.com/fluxcd/flux2-multi-tenancy)
 - [Migrate from Helm Operator to Flux v2](/docs/migration/helm-operator-migration/)
 - [How to structure your HelmReleases](https://github.com/fluxcd/flux2-kustomize-helm-example)
+
+## Release questions
+
+### Where can I find information on how long versions are supported and how often to expect releases?
+
+Releases are as needed for now and a release page with the schedule will be created when Flux reaches general availability status. We understand the importance of having a release schedule and are aware that such a schedule will support the users' ability to plan for an upgrade in advance.
```
dholbach, line 393:
```suggestion
Releases are as needed for now and a release page with the schedule will be created when Flux reaches general availability status. We understand the importance of having a release schedule and are aware that such a schedule will support the users' ability to plan for an upgrade in advance.

```

dholbach: Maybe point to `/roadmap` for now as well to at least point to some available information?

Maybe also point out that `flux2` in 2021 had at least monthly releases, often more and all controllers together 100+ releases in that time-frame? That at least would demonstrate some of the release velocity and while not a strict guarantee - at least an indication of progress (even if in a "pre-release" kind of fashion still).

stefanprodan: We can say that the release cadence is at least once per month.

priyanka-ravi: Let me know if the update looks good! If y'all think I should add in more info about our current releases let me know! Thanks!!

scottrigby (approved):
👍

scottrigby:
The linkchecker was failing on an unrelated file, so I merged this. We should look into that. I tried running it twice. Example failing run: https://github.com/fluxcd/website/runs/4972626006?check_suite_focus=true#step:9:56

Issue #587, opened 2021-10-18
Blurry line between Flux and GitOps Toolkit


dholbach:
# Blurry line between “GitOps toolkit” and “Flux”

> [..] when going through the documentation
> (https://fluxcd.io/docs) sequentially, then the first mention of “GitOps toolkit” is in the second
> page within the “Guides” section, named “Helm Releases”:
> https://fluxcd.io/docs/guides/helmreleases/
> On this page it says a prerequisite to the guide is to have a Kubernetes cluster with the
> GitOps controllers installed and then referencing the getting started and installation guides,
> as shown by the following Figure.

![grafik](https://user-images.githubusercontent.com/1346979/137730734-880e173c-0fcb-46df-8b09-a0c096141faa.png)

> However, “GitOps toolkit” is not mentioned on these pages.
> In addition to the above observations, the current destination of the link to the gitops toolkit
> on the fluxcd.io page, namely https://fluxcd.io/#gitops-toolkit, does not lead anywhere. This is
> the link on the rightmost box on the website as shown by the following Figure:

![grafik](https://user-images.githubusercontent.com/1346979/137730794-6cb055c1-c636-4b9b-9433-02c6b4b8ab6a.png)

##  Recommendation

> Flux uses names and concepts conservatively to easily convey the structure of the
> infrastructure. However, the documentation must have an equal structure including proper
> introduction of concepts. We recommend using clear naming of the components in the
> system and ensuring proper introduction without broken links or links to pages that do not
> clarify concepts that were promised. It would be a great assistance to have a page that
> unifies the nomenclature of Flux and links to details about the definition or where the concept
> is introduced.

priyanka-ravi:
I can attempt to tackle this one!

scottrigby:
Re this part:
> It would be a great assistance to have a page that unifies the nomenclature of Flux and links to details about the definition or where the concept is introduced.

I pointed @priyanka-ravi to this PR, which I had still needed to update. Now updated  to resolve @squaremo's later suggestions. 

@priyanka-ravi  is reviewing now with this issue above in mind. This is essentially a unified page clearly disambiguating nomenclature about the various things we refer to when saying “Flux”, how it relationship to GitOps Toolkit, etc. Once reviewed again, we can update website automation to pull it somewhere into the docs.

dholbach:
[...]

PR #728, opened 2022-02-01
De-emphasise GitOps Toolkit somewhat

dholbach:
On the typical path of a Flux user, they will read "GitOps Toolkit" a lot. In this PR I try to collect GOTK information more centrally, so that the flow of a Flux user ("I want to 	understand/install Flux") is interrupted less.

Also fix link on landing page - it was broken.

Re-use phrasing from https://github.com/fluxcd/community/blob/main/style-guide.md

Fixes: #587 

cc @priyanka-ravi


scottrigby:
Linkchecker is broken, but links work. Merging!

PR #735, opened 2022-02-02
Update GA roadmap

stefanprodan:
As of v0.26.0, platform admins have the option to [lock down Flux](https://fluxcd.io/docs/installation/#multi-tenancy-lockdown) on multi-tenant clusters.

Given that now we offer a way to isolate tenants at namespace level, I propose we move RFC-0002 and RFC-0003 out of the GA requirements. These RFCs would improve Flux UX in regard to tenant onboarding and resources sharing across the cluster, but the proposed improvements are no longer a hard requirement for reaching general availability.

### DIFF
--- a/content/en/roadmap.md
+++ b/content/en/roadmap.md
@@ -49,14 +49,14 @@ this list is subject to change while we gather feedback:
     - [x] Replace `kubectl` usage in Flux CLI with server-side apply
     - [x] Rewrite the kustomize-controller reconciler using server-side apply
 
+- Multi-tenancy lockdown
+    - [x] [Allow setting a default service account for impersonation](https://github.com/fluxcd/flux2/issues/2340)
+    - [x] [Allow disabling cross-namespace references](https://github.com/fluxcd/flux2/issues/2337)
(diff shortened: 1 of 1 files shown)


dholbach:
We discussed this in [our meeting today](https://docs.google.com/document/d/1l_M0om0qUEN_NNiGgpqJ2tvsF2iioHkaARDeh6b70B0/edit#heading=h.495nesq4hd47). If folks have feedback, please comment here, on Slack or in the [meeting next Thursday](https://fluxcd.io/#calendar)?

scottrigby (approved):
I was not in the early Flux meeting, but given the status and priority, this makes sense and is good to clarify to others following the road to GA.

darkowlzz (approved):
SGTM!

relu (approved):
Makes sense to me 👍

hiddeco (approved):

dholbach (approved):
What do you think about communications regarding "here's the last things on our list before GA" and "here's how we are planning to enhance our security story - we want your input".. or something along those lines?

makkes (approved):

PR #750, opened 2022-02-09
Fix bad indentation of Kustomize patch examples

davivcgarcia:
The examples provided are with bad indentation for the Kustomize patch, generating an reconciliation error when applied like below:

```
$ kustomize build
Error: trouble configuring builtin PatchTransformer with config: `
patch: "- op: add\n path: /spec/template/spec/containers/0/args/-\n value: --aws-autologin-for-ecr
  \  "
target:
  group: apps
  kind: Deployment
  name: image-reflector-controller
  namespace: flux-system
  version: v1
`: unable to parse SM or JSON patch from [- op: add
 path: /spec/template/spec/containers/0/args/-
 value: --aws-autologin-for-ecr]
```

### DIFF
--- a/content/en/docs/guides/image-update.md
+++ b/content/en/docs/guides/image-update.md
@@ -665,8 +665,8 @@ patches:
     namespace: flux-system
   patch: |-
     - op: add
-     path: /spec/template/spec/containers/0/args/-
-     value: --aws-autologin-for-ecr
+      path: /spec/template/spec/containers/0/args/-
+      value: --aws-autologin-for-ecr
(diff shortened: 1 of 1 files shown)


kingdonb:
Thanks for the contribution @davivcgarcia 🎉 

Can you please `--sign-off` your commit to pass the failing CI for DCO? Check the instructions at https://github.com/fluxcd/website/pull/750/checks?check_run_id=5127029638 for help with this.

dholbach (approved):

PR #770, opened 2022-02-16
Remove e2e docs section 6 merge conflict

sbernheim:
Removing a git merge conflict apparently committed and pushed in error.

@kingdonb please review and correct if this is not what you were going for....


kingdonb:
I amended your commit through a rebase, the only difference between what you had written and how it reads now is that I've increased the emphasis and separated from other information the fact that GitRepository source as a service is completely read-only. (If I've done the build-up correctly, it will be important that readers have internalized this information when they get to the section on Helm Chart+GitRepository and "the things you will need to be aware of" when using GitRepo that way.)

kingdonb (approved):
LGTM 👍

PR #771, opened 2022-02-16
Refine server-side apply sec 8

sbernheim:
Refinements for section 8 on server-side apply.

There does not appear to be a detailed description of server-side apply in the FluxCD documentation, so I've provided links to the Kubernetes documentation on server-side apply.


kingdonb:
I am not sure if this is the PR where these got mixed up, or if it happened earlier, but SOPS decryption happens before Kustomize Build, and this ought to be important since `substituteFrom` may reference secrets, (although I'm not certain it has the effect I think it does here...)

Anyway, LGTM 👍  I'll be merging it and then giving one long final proofread of the whole doc, while I extract links and pull them out of the text and down to the bottom footer. Then it should be marked as ready for review, after a few more minor changes.

kingdonb (approved):
LGTM 👍  merging back to my WIP branch for further iterating

PR #776, opened 2022-02-17
Add Evrone to adopters

akirill0v:
Signed-off-by: Aleksandr Kirillov <kirillov@evrone.com>

Our company (Evrone.com) has been using  Flux(v1 and v2) almost for for 2 years. We use it on our internal products and setup customer Kubernetes environments with it. It would be great to be on the adopters list. Thanks )


stefanprodan (approved):
Thank you @akirill0v 🥇

Issue #720, opened 2022-01-25
Discuss where content is coming from in main README


dholbach:
From the CNCF Tech Docs assessment of our docs (cncf/techdocs#89):

> - Put [this information](https://github.com/fluxcd/website/blob/main/content/en/docs/contributing/docs/some-background.md#running-the-site-locally) directly in the README.md of the website repository, because people are lazy. Kudos on trying to keep everything in one place, however!

PR #793, opened 2022-02-22
Move 'running the site' info to main README.md

dholbach:
Feedback from the CNCF Techdocs assessment (/cncf/techdocs#89) was that this is the most natural place for this kind of information.

Fixes: #720


scottrigby (approved):
👍

PR #795, opened 2022-02-22
Add tooltip on the adopters' logos

darkowlzz:
Some logos don't reveal the name of the adopter. Add tooltip to show the
name when hovered.

tooltip


dholbach (approved):
This is great - thanks @darkowlzz

PR #811, opened 2022-03-03
Change BlueStone to Netrics as adopter

janlauber:
As announced in the following post of Netrics, BlueStone got bought by Netrics:
https://www.netrics.ch/bluestone-wird-netrics

So the adopters page of flux must change.

Thanks for merging

Signed-off-by: Jan Lauber <jan.lauber@protonmail.ch>


stefanprodan (approved):
Thanks for the update @janlauber

PR #818, opened 2022-03-08
Fix highlight in "Customize Flux manifests"

michaelsatish:
Change in line 528. `kustomization.yam`l to `kustomization.yaml`


dholbach (approved):
Thanks a lot for the fix @michaelsatish !

PR #789, opened 2022-02-21
update openshift instructions

chanwit:
Signed-off-by: Chanwit Kaewkasi <chanwit@gmail.com>

PR #822, opened 2022-03-09
Add blog entry about pod security standard 'restricted'

dholbach:

### DIFF
--- a/content/en/blog/2022-03-09-pod-security-standard-restricted.md
+++ b/content/en/blog/2022-03-09-pod-security-standard-restricted.md
@@ -1,0 +1,151 @@
+---
+author: dholbach
+date: 2022-03-09 12:30:00+00:00
+title: 'Security: Using Pod Security Standard "restricted"'
+description: pod security standards is a recent addition to Kubernetes, coming to replace pod security policies. Alongside seccomp, it provides greater isolation levels to workloads. Read up on how we moved all Flux controllers to 'restricted' mode and how that's going to keep you safer.
+url: /blog/2022/03/security-pod-security-standard-restricted/
+tags: [security]
+---
+
+
+Next up in our [blog series about Flux
+Security](/tags/security/) is how we moved
+to Pod Security Standard "restricted", all the background info you need
+to know and how that makes things safer for you.
+
+[Since version 0.26 of
+Flux](/blog/2022/01/january-update/#security-news)
+we are applying
+
+> \[..\] the [restricted pod security
+> standard](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted)
+> to all controllers. In practice this means:
+>
+> - all Linux capabilities were dropped
+> - the root filesystem was set to read-only
+> - the `seccomp` profile was set to the runtime default
+> - run as non-root was enabled
+> - the filesystem group was set to 1337
+> - the user and group ID was set to 65534
+>
+> Flux also enables the Seccomp runtime default across all controllers.
+> Why is this important? Well, the default `seccomp` profile blocks key
+> system calls that can be used maliciously, for example to break out of
+> the container isolation. The recently disclosed [kernel vulnerability
+> CVE-2022-0185](https://blog.aquasec.com/cve-2022-0185-linux-kernel-container-escape-in-kubernetes)
+> is a good example of that.
+
+## Pod Security Standards definition
+
+Kubernetes defined three policies in its Pod Security Standards. They
+range from
+
+- **Privileged**: This does not place any restrictions on the workload
+  at all. The idea being that this can be used for system- and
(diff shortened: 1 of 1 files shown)


dholbach:
→ https://deploy-preview-822--fluxcd.netlify.app/blog/

content/en/blog/2022-03-09-pod-security-standard-restricted.md
```diff
@@ -0,0 +1,125 @@
...
+specified.
+
+In its first inception `seccomp` was introduced into Linux in 2005, to
+Docker in version 1.10 (Feb 2016) and to Kubernetes in version 1.3 (Jul
+2016). So while the technology has been around for a while and you could
+handcraft your own `seccomp` profiles, the challenge has always been
+striking the right balance: if you are too generous in your filter, it
+won't guard against malware effectively -- if you are too strict, your
+application might not work.
+
```
pjbgf, line 84: It may be worth pointing out that as of `v1.24` Kubernetes still runs all workloads with seccomp in `unconfined` mode, in other words, disabled. On the other hand, docker has seccomp enabled by default for years now - I believe it was soon after they started supporting it.

There are plans to change the Kubernetes default on `v1.25`, and have all workloads set to `RuntimeDefault` unless opted-out - this is yet to be confirmed. This will be based on `SeccompDefault` [feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) being enabled from that version onwards.

content/en/blog/2022-03-09-pod-security-standard-restricted.md
```diff
@@ -0,0 +1,125 @@
...
+```yaml
+spec:
+  securityContext:
+    seccompProfile:
+      type: RuntimeDefault
+```
+
+All Flux controllers have this implemented as well now! This is just one
+more measure we are taking to keep you more secure.
```
pjbgf, line 99:
```suggestion
All Flux controllers have this enforced now - regardless of the cluster-level settings.
This is just one more measure we are taking to keep you more secure.
```

content/en/blog/2022-03-09-pod-security-standard-restricted.md
```diff
@@ -0,0 +1,125 @@
...
+
```
pjbgf, line 100: Wrapping up we could just summarise that by adopting both changes, we further restrict the permissions that flux requires in order to operate. This, alongside other changes we are working on, translates in a decreased attack surface which may reduce the impact of eventual CVEs that may surfaced in our code base - or our supply chain.

pjbgf (commented):
@dholbach really good stuff! Added a few comments, but as always it is looking really good.

stefanprodan:
I would tell people that due to this change Flux no longer works on OpenShift, maybe link to @pjbgf issue.

dholbach:
> I would tell people that due to this change Flux no longer works on OpenShift, maybe link to @pjbgf issue.

Hum... wasn't this addressed through https://github.com/fluxcd/website/pull/789?

dholbach:
Ah OK, I guess you are referring to https://github.com/fluxcd/source-controller/issues/582 and https://github.com/openshift/cluster-kube-apiserver-operator/issues/1325? I might go and add a note pointing to this.

stefanprodan:
> Ah OK, I guess you are referring to https://github.com/fluxcd/source-controller/issues/582 and  https://github.com/openshift/cluster-kube-apiserver-operator/issues/1325? I might go and add a note pointing to this.

Yes maybe @pjbgf can help with a short note on this. FYI The Flux published on OperatorHub by Weaveworks removed the profile to not break the one-click install https://github.com/weaveworks/flux2-openshift/blob/main/kustomization.yaml#L19

content/en/blog/2022-03-09-pod-security-standard-restricted.md
```diff
@@ -0,0 +1,151 @@
...
+{{% note %}}
+If you are an OpenShift user, you might run into [this
+issue](https://github.com/fluxcd/source-controller/issues/582)
+([related upstream
+report](https://github.com/openshift/cluster-kube-apiserver-operator/issues/1325)).
+The work-around right now is to remove the seccomp profile as
+described in [these instructions](/docs/use-cases/openshift/#flux-installation-with-cli).
+{{% /note %}}
+
```
dholbach, line 88: @pjbgf WDYT?

pjbgf: Yes, @dholbach I like it - precise and to the point, whilst providing all the info the user may need. :+1:

stefanprodan (approved):
LGTM

Thanks @dholbach

pjbgf (approved):
LGTM - great job @dholbach!

Issue #717, opened 2022-01-25
Sections "Guides" and "Use cases" overlap


dholbach:
From the CNCF Tech Docs assessment of our docs (https://github.com/cncf/techdocs/pull/89):

> Between the [Guides](https://fluxcd.io/docs/guides/) and [Use cases](https://fluxcd.io/docs/use-cases/) overlap a bit. The topics in these two sections are around installing and using Flux with a variety of different technologies. The names of both are somewhat vague which causes a lack of clarity for the reader: they need to click in to understand what each section is about, and even after doing so it isn't entirely clear.
>
> ### Recommendations
> 
> **Information architecture**:
> 
> - Reorganize the [Guides](https://fluxcd.io/docs/guides/) and [Use cases](https://fluxcd.io/docs/use-cases/): 
>     - Docs root
>         - ...
>         - Installation
>         - Integrating Flux
>             - Azure
>             - OpenShift
>             - Helm 
>             - Jenkins 
>             - Prometheus
>             - Flux Notifications
>         - Automation & GitOps 
>             - Ways of structuring your repositories
>             - Setup webhook receivers
>             - Sealed Secrets
>             - Manage Kubernetes secrets with Mozilla SOPS 
>             - Automate image updates to Git
>             - How to make sortable image tags to use with automation 
>             - GitHub Actions Manifest Generation
>             - GitHub Actions Auto Pull Request

dholbach:
Interesting point in https://github.com/fluxcd/website/issues/740#issuecomment-1031656530

> I think it would be helpful if the Flux docs more clearly outlined day 1/day 2 operations, the GitOps workflow, and how Renovate fits into that. Right now, the [Guides](https://fluxcd.io/docs/guides/) and the [Use Cases](https://fluxcd.io/docs/use-cases/) sections are like cookbooks with a mish-mash of various recipes, which is a little bit confusing. I'd suggest reorganizing the contents of those sections into two new sections: Installation/Configuration (i.e. day 1) and Operations (i.e. day 2). Then I'd add an Operations subsection to discuss manual and automated upgrades via flux bootstrap, the GitHub Action, Renovate, etc. I'd probably put the Prometheus monitoring guide and migrations bits in the Operations section too.

danports:
Yes, exactly, I was thinking about writing up a suggested docs outline, but it sounds like I was beaten to the punch. 😂 I think that outline could still use some improvement, though...I'll give it some thought.

danports:
[...]

PR #845, opened 2022-03-21
Reorg docs IA

dholbach:
Ref: #717

This is likely going to be a draft PR for a while. For now I'd like to make this PR represent a simple PoC based off of https://github.com/fluxcd/website/issues/717#issuecomment-1034365856 - so no big moving of docs just yet. More of a place to try a couple of things and review the information architecture (IA) together.


dholbach:
Bildschirmfoto von 2022-03-23 11-37-54

dholbach:
The above represents the simple changes for now. So essentially just navbar changes including redirects. Anchors should continue to work. No content moved.

What does everyone think so far?

kingdonb:
I don't see any problems and in general I think this is more accessible 🎉 

I've mainly reviewed the TOC that you posted earlier, it seems like a reasonable organization structure. It seems like fewer top-level entries, but after counting I'm pretty sure both versions have 13 top-level entries. There are things nested more deeply and it's clearer when you need to access each doc. I think it will be less difficult to find what you're looking for, but I will try using the preview version for a little while and have to see how I feel after that...

scottrigby:
My only concern is that right now "platforms" appears only under "installation", but the use cases for specific platforms we have now apply to more than just installation 🤔 Should there be a top level platforms section, which summarizes info for each major platform, and links to the other docs as appropriate (under installation > platforms, operations > platforms, workflows > platforms)? I know people will be looking for this.

dholbach:
This is based off of https://github.com/fluxcd/website/issues/717#issuecomment-1034365856 - which to me seemed to bring more clarity from a nav and organisational point of view. If you have other suggestions on what to place where, please let me know.

What we could (and should) do separately is apart from the navigation: make better use of the /docs/ and /docs/installation landing pages more and build a clearer flow for the reader. For example could we place project/platform logos in the docs to get the reader's attention and direct them to the right place.

dholbach:
Feedback from @wwentland

# Organisation

There is a lot of very good content and I personally consider the documentation to be very helpful, if a little hard to navigate due to its organically grown nature.

## Proposal by @danports in <https://github.com/fluxcd/website/issues/717#issuecomment-1034365856>

 - Nice separation of content into different sections focused on user/reader intent
 - I think this would benefit from standardising on either “Noun” or “Verb” section titles (where appropriate)
  - Soft personal preference on “Verb” ones, but consistency most important
  - “Installation” vs “Installing” / “Integrating” vs “Integration”
- We might want to introduce a “Reference” section, subsuming CLI and Components reference style documentation
- I specifically like the separation in Getting started (day 1), and typical operations (day 2) content
  - Day 2 / Operations section would be perfect for discussing 2.0 HA topics also

## This proposal

- Nice adaptation of the earlier proposal with the goal of not introducing too many changes to the underlying content/keeping anchors working
- Migration from Flux v1 too prominent IMHO (increasingly less important)
- Some sections assume prior knowledge
  - “Flagger” would be better as “Progressive Delivery” or “Progressive Delivery with Flagger”
  - Toolkit Components  would be much better as part of the “Workflow” or “How to build custom things w/ Flux” section
  - Toolkit Dev Guide would fit much better in either a “Flux as SDK” or “Integrating” section

Issue #797, opened 2022-02-23
Configure image automation authentication - containercannotrun issue


js-johnpope:
I am new to flux and working my way through an install of it.  I already have an application managed by flux and am just adding the image automation piece.

I have not used it extensively and may be missing something here, but I was pretty stuck on this for a day.

Using the link [here](https://fluxcd.io/docs/guides/cron-job-image-auth/#using-cronjob-to-sync-ecr-credentials-as-a-kubernetes-secret) to configure the ECR authentication to enable sync against my image repo I was getting a 'ContainerCannotRun' error for the 'create-secret' container.  

I already set up my IAM role with the correct access as per the IRSA requirement and am using that in the cronjob so was sure it was not that.

When I describe the failed job, there is a failed message like `cannot stat '/bin/bash'`.  Thus the container is not even starting to be able to run its command to create the secret.

I tried running the flux-cli:v0.25.2 image locally trying to use and see that there is indeed no /bin/bash:

![Screenshot 2022-02-23 at 11 59 12](https://user-images.githubusercontent.com/43172431/155315333-504caa60-9f12-417f-b076-4e03d453b4d1.png)
 
Even if the flux-cli container did work, there is no kubectl command (I think).  

I swapped out the flux-cli image to a custom one I use that has kubectl installed and the secret got created and the scan of the repo worked just fine.

So it feels like:

1. The flux-cli image is not needed here or the secret creation command is incorrect
2. A kubectl image is needed
3. I am missing something...

Setup:

Flux v2
Kubernetes v1.19
flux-cli:v0.25.2

Thanks, J

stefanprodan:
kubectl is included in the flux-cli image https://github.com/fluxcd/flux2/blob/main/Dockerfile, we need to change the command to `/bin/sh`

js-johnpope:
Using a custom image with kubectl worked for me, the documented process did not.

js-johnpope:
Re-opening.

The fix suggested by @morganchristiansson looks good?

js-johnpope:
https://github.com/fluxcd/website/pull/846

PR #846, opened 2022-03-22
Fix comands in ecr-credentials-sync CronJob

morganchristiansson:
`bash` is not available in docker image and `$(<` doesn't work with `sh`

Fix: #797

### DIFF
--- a/content/en/docs/guides/cron-job-image-auth.md
+++ b/content/en/docs/guides/cron-job-image-auth.md
@@ -115,14 +115,14 @@ spec:
             - mountPath: /token
               name: token
             command:
-            - /bin/bash
+            - /bin/sh
             - -ce
             - |-
               kubectl create secret docker-registry $SECRET_NAME \
                 --dry-run=client \
                 --docker-server="$ECR_REGISTRY" \
                 --docker-username=AWS \
-                --docker-password="$(</token/ecr-token)" \
+                --docker-password="$(cat /token/ecr-token)" \
                 -o yaml | kubectl apply -f -
 ```
 


content/en/docs/guides/cron-job-image-auth.md
```diff
@@ -115,14 +115,14 @@ spec:
...
-            - /bin/bash
+            - /bin/sh
             - -ce
             - |-
               kubectl create secret docker-registry $SECRET_NAME \
                 --dry-run=client \
                 --docker-server="$ECR_REGISTRY" \
                 --docker-username=AWS \
-                --docker-password="$(</token/ecr-token)" \
+                --docker-password="$(<cat token/ecr-token)" \
```
stefanprodan, line 125: I think you need to remove `<` 
```suggestion
                --docker-password="$(cat token/ecr-token)" \
```

stefanprodan:
@morganchristiansson can you confirm you tested the above on a real cluster? Also please signoff your commit using the Git CLI, see here how to do it https://github.com/fluxcd/website/pull/846/checks?check_run_id=5641691069

content/en/docs/guides/cron-job-image-auth.md
morganchristiansson: Thanks! Quickly spotted! Amended commit and force pushed.

morganchristiansson:
We just did flux v2 migration and deployed ecr-credentials-sync. After replacing `bash` with `sh` the job was failing with error because `--docker-password=` was provided empty when using `sh` with `$(<`.

This was previously reported in https://github.com/fluxcd/website/issues/797 by my colleague

morganchristiansson:
This is verifiable using `docker` command

```
$ docker run --rm -ti --entrypoint=sh ghcr.io/fluxcd/flux-cli:v0.25.2 
/ # echo test > test
/ # echo $(< test)

/ # echo $(cat test)
test
```

stefanprodan (approved):
LGTM

Thanks @morganchristiansson

PR #859, opened 2022-04-01
Add note to image-update.md about namespaces

jiphex:
When an ImageUpdateAutomation is specified in namespace A, the image-automation controller will select imagepolicy setters which match the namespace of the ImageUpdateAutomation object

This caused me some confusion for a while - in our case we have a single `flux-system` ImageUpdateAutomation resource in our bootstrapped flux repositories, and because this wasn't explicitly mentioned at this point in the docs, I assumed it would make more sense to put the ImagePolicy and ImageRepository resources in the namespaces of their relevant deployments, rather than in the `flux-system` one, however these aren't found by the image-automation setters.

The language could be improved, but I was struggling to word the explanation.


staceypotter:
@jiphex can you sign off your commit please? thx

PR #862, opened 2022-04-01
Mention Artifact Registry in image update docs

kingdonb:
We support both GCR (legacy) and Google Artifact Registry, the newer alternative

https://cloud.google.com/blog/products/application-development/understanding-artifact-registry-vs-container-registry

Should we include this link in the docs as well, I think it is not needed but maybe could be helpful as a reference too.


stefanprodan (approved):
LGTM

Thanks @kingdonb 🏅

PR #877, opened 2022-04-09
changed flux-system to app1

lakamsani:
Around line 167, In step 1 of moving a file from ks `app1`  to ks `app2` it appears `flux-system` was accidentally used instead of `app1`?


yebyen:
I don't think this is an error.

The app1 and app2 Kustomization resources are both defined in the flux-system namespace, so that is the one that needs to be reconciled in order to be sure that the definition is made fresh from the source.

PR #901, opened 2022-04-27
Move rungway from legacy to v2 adopters

endrec:
Signed-off-by: Endre Czirbesz <endre.czirbesz@rungway.com>


stefanprodan (approved):
Thank you @endrec 🏅

PR #912, opened 2022-05-03
Add the deploy key rotation procedure to install docs

stefanprodan:
Solves https://github.com/fluxcd/flux2/discussions/2694


pjbgf (approved):
LGTM!

Issue #913, opened 2022-05-03
Blog: videos in youtube shortcode too big


dholbach:
Check out https://fluxcd.io/blog/2021/11/december-update/#michael-irwin---compose-with-k8s-and-flux

The size of the videos shown is too big.

dholbach:
So it looks like in the [youtube shortcode](https://gohugo.io/content-management/shortcodes/#youtube) we can override `class` - can somebody please help figure out how to set it right?

PR #915, opened 2022-05-03
update youtube shortcode

Arhell:
resolve https://github.com/fluxcd/website/issues/913

### DIFF
--- a/layouts/shortcodes/youtube.html
+++ b/layouts/shortcodes/youtube.html
@@ -1,0 +1,9 @@
+<div class="responsive-video">
+  <iframe
+    src="https://www.youtube.com/embed/{{ index .Params 0 }}"
+    allowfullscreen
+    frameborder="0"
+    title="video"
+  >
+  </iframe>
+</div>
--- a/layouts/shortcodes/youtube.html
+++ b/layouts/shortcodes/youtube.html
@@ -1,9 +1,10 @@
-<div class="responsive-video">
-  <iframe
-    src="https://www.youtube.com/embed/{{ index .Params 0 }}"
-    allowfullscreen
-    frameborder="0"
-    title="video"
-  >
-  </iframe>
+{{- $pc := .Page.Site.Config.Privacy.YouTube -}}
+{{- if not $pc.Disable -}}
+{{- $ytHost := cond $pc.PrivacyEnhanced  "www.youtube-nocookie.com" "www.youtube.com" -}}
+{{- $id := .Get "id" | default (.Get 0) -}}
+{{- $class := .Get "class" | default "responsive-video" -}}
+{{- $title := .Get "title" | default "YouTube Video" }}
+<div class="{{ $class }}">
+  <iframe src="https://{{ $ytHost }}/embed/{{ $id }}{{ with .Get "autoplay" }}{{ if eq . "true" }}?autoplay=1{{ end }}{{ end }}" {{ if not $class }}style="position: absolute; top: 0; left: 0; width: 100%; height: 100%; border:0;" {{ end }}allowfullscreen title="{{ $title }}"></iframe>
 </div>
+{{ end -}}
--- a/assets/scss/_responsive-video.scss
+++ b/assets/scss/_responsive-video.scss
@@ -1,0 +1,20 @@
+.responsive-video {
+  position: relative;
+  overflow: hidden;
+  width: 100%;
+  padding-top: 56.25%;
+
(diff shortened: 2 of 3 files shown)


Arhell:
cc @dholbach

Arhell:
https://deploy-preview-915--fluxcd.netlify.app/blog/2021/11/december-update/#michael-irwin---compose-with-k8s-and-flux

layouts/shortcodes/youtube.html
```diff
@@ -0,0 +1,8 @@
+<div class="responsive-video">
+  <iframe
+    src="https://www.youtube.com/embed/{{ index .Params 0 }}"
+    allowfullscreen
+    frameborder="0"
+  >
+  </iframe>
+</div>
```
dholbach, line 8: Any chance, we can get this closer to https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/shortcodes/youtube.html which is the current `youtube` shortcode we use?

Or can we maybe just use `{{< youtube id="abcabcba" "class"="responsive-video" >}}` now without introducing an override for the `youtube` shortcode?

Arhell: I think that what I did is a universal option, it will not require you to change anything in the documentation. In any case, it will be possible to resize or something else in the iframe

or make a copy with link and fix it for our needs?

dholbach: How about we add a new shortcode called `yt` that does this:

```
{{ $params := .Get 0 }}
{{ partial "youtube" (dict (class "responsive-video") (id $param1) }}
```

and move all occurrences to use it?

That way we would still benefit from everything that's defined in the hugo shortcode itself.

Arhell: ok

layouts/shortcodes/yt.html
```diff
@@ -0,0 +1,12 @@
...
+
+<div {{ with .Get "class" }} class="{{ . }}"{{ end }}>
+  <iframe
+    src="https://www.youtube.com/embed/{{ index .Params 0 }}"
+    allowfullscreen
+    frameborder="0"
+  >
+  </iframe>
+</div>
+{{- end }}
```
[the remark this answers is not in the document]
Arhell: I tried to add the class in different ways, but alas, it did not work out.
https://gohugo.io/functions/dict/#readout

layouts/shortcodes/youtube.html
dholbach: Ok - I thought I had this working locally - unfortunately it doesn't.

layouts/shortcodes/youtube.html
```diff
@@ -0,0 +1,9 @@
+<div class="responsive-video">
```
dholbach, line 1: Your solution works great - thanks a lot for that!

I'm just wondering if we should model this shortcode override more closely on https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/shortcodes/youtube.html so that folks can still use the options as documented...?


```diff
--- /home/daniel/dev/website/hugo/tpl/tplimpl/embedded/templates/shortcodes/youtube.html	2021-09-13 13:58:27.103370314 +0200
+++ layouts/shortcodes/youtube.html	2022-05-16 11:19:33.660974736 +0200
@@ -2,9 +2,9 @@
 {{- if not $pc.Disable -}}
 {{- $ytHost := cond $pc.PrivacyEnhanced  "www.youtube-nocookie.com" "www.youtube.com" -}}
 {{- $id := .Get "id" | default (.Get 0) -}}
-{{- $class := .Get "class" | default (.Get 1) -}}
+{{- $class := .Get "class" | default "responsive-video" -}}
 {{- $title := .Get "title" | default "YouTube Video" }}
-<div {{ with $class }}class="{{ . }}"{{ else }}style="position: relative; padding-bottom: 56.25%; height: 0; overflow: hidden;"{{ end }}>
+<div class="{{ $class }}">
   <iframe src="https://{{ $ytHost }}/embed/{{ $id }}{{ with .Get "autoplay" }}{{ if eq . "true" }}?autoplay=1{{ end }}{{ end }}" {{ if not $class }}style="position: absolute; top: 0; left: 0; width: 100%; height: 100%; border:0;" {{ end }}allowfullscreen title="{{ $title }}"></iframe>
 </div>
 {{ end -}}
```

Not sure if anyone else feels strongly about this...?

PR #916, opened 2022-05-04
Update bootstrap docs

stefanprodan:
Changes:
- Add Helm repository index caching to bootstrap cheatsheet
- Disable Kustomize remote bases on multi-tenant clusters

### DIFF
--- a/content/en/docs/cheatsheets/bootstrap.md
+++ b/content/en/docs/cheatsheets/bootstrap.md
@@ -357,8 +357,8 @@ patches:
 
 ### Multi-tenancy lockdown
 
-Lock down Flux on a multi-tenant cluster by disabling cross-namespace references
-and setting a default service account:
+Lock down Flux on a multi-tenant cluster by disabling cross-namespace references and Kustomize remote bases, and
+by setting a default service account:
 
 ```yaml
 apiVersion: kustomize.config.k8s.io/v1beta1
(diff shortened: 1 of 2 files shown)


content/en/docs/cheatsheets/bootstrap.md
```diff
@@ -193,6 +193,33 @@ patches:
...
+        path: /spec/template/spec/containers/0/args/-
+        value: --helm-cache-ttl=60m
+      - op: add
+        path: /spec/template/spec/containers/0/args/-
+        value: --helm-cache-purge-interval=5m
+    target:
+      kind: Deployment
+      name: source-controller
+```
+
```
hiddeco, line 222:
```suggestion

When `helm-cache-max-size` is reached, an error is logged and the index is instead
read from file. Cache hits are exposed via the `gotk_cache_events_total` Prometheus
metrics. Use this data to fine-tune the configuration flags.
```

hiddeco (commented):
We should maybe also add an entry about QPS and burst flags, especially for high traffic helm-controller instances, these values should be configured to higher values (e.g. `500/800` or even `800/1000`). With a note this may increase CPU and other resource usage however.

stefanprodan:
@hiddeco the QPS is already covered, see https://fluxcd.io/docs/cheatsheets/bootstrap/#increase-the-number-of-workers

hiddeco (approved):

PR #918, opened 2022-05-05
Add Grafana Labs as adopters of Flux V2 and Flagger

julienduchesne:
Signed-off-by: Julien Duchesne <julien.duchesne@grafana.com>


stefanprodan (approved):
Thanks @julienduchesne 🤗

mewzherder:
Hi @julienduchesne Thanks so much for taking the time to do this. Would you be willing to chat with us about your Flux experience? We'd like to learn more about use cases and gather feedback. Please let me know. Thanks!

PR #937, opened 2022-05-13
use old search widget - can be dropped once we merge #931

### LINKED ISSUE #931: Transition to Algolia v3
Unfortunately we had to revert #929 in #930 because the search for tag `legacy` vs `current` did not work.

This is likely a mistake in how our crawler config was transitioned by the Algolia team. I'll reach out and ask if they can help us fix it.

`algoliaOptions` also needs to be changed to `searchParameters` in the new API.

dholbach:
The last commit broke search. Testing locally I didn't see this coming. Sorry about this.

Once we resolve #931, we can drop this search override.


kingdonb (approved):
LGTM

PR #960, opened 2022-05-26
Add Cookpad as an adopter of Flux V2

mattjw:
I came here via the call-for-adopters shared on the CNCF Slack during Kubecon EU. We've been using Flux at Cookpad for a while (~2 years, if you count both Flux v1 and v2) in a number of production clusters. 🎊 

P.S. It was great to chat to Flux users and maintainers in Valencia last week ❤️.

Signed-off-by: Matt J Williams <matt-williams@cookpad.com>


stefanprodan (approved):
LGTM

Thank you @mattjw 🏅

PR #982, opened 2022-06-07
Document the controllers flags

stefanprodan:
Fix: https://github.com/fluxcd/flux2/issues/2654


pjbgf (approved):
Great job @stefanprodan! :clap::clap:

LGTM

PR #986, opened 2022-06-09
helm gh actions guide: Exclude events related to dependencies check

stefanprodan:
Avoid promotion for "Dependencies do not meet ready condition" events.


dholbach (approved):
Thanks

PR #990, opened 2022-06-09
Add documentation for how to run jobs with Flux

phillebaba:
Adds documentation for how to run Jobs with Flux with depending Kustomizations.

### DIFF
--- a/content/en/docs/use-cases/running-jobs.md
+++ b/content/en/docs/use-cases/running-jobs.md
@@ -1,73 +1,142 @@
 ---
-title: Running Jobs With Flux
-linkTitle: Running Jobs
-description: "How to run Jobs with Flux."
+title: Running pre and post-deployment jobs with Flux
+linkTitle: Running Kubernetes Jobs with Flux
+description: "How to run Kubernetes Jobs before and after an application deployment with Flux."
 weight: 10
 ---
 
-Additional considerations have to be made when managing Kubernetes Jobs with Flux. By default if you were to have Flux manage a single Job resource, it would apply it once to the cluster. The Job would trigger a run which would create a Pod that can either error or run to completion. Attempting to update the Job after it has been applied to the cluster will not be allowed as changed to the Jobs `spec.Completions`, `spec.Selector` or `spec.Template` is not permitted by the Kubernetes API. To be able to update a Job the Job has to be recreated by first beeing removed and the reapplied to the cluster.
+Additional considerations have to be made when managing Kubernetes Jobs with Flux.
+By default, if you were to have Flux reconcile a Job resource,
+it would apply it once to the cluster, the Job would create a Pod that can either error or run to completion.
+Attempting to update the Job manifest after it has been applied to the cluster is not be allowed, as changes to the
+Jobs `spec.Completions`, `spec.Selector` and `spec.Template` are not permitted by the Kubernetes API.
+To be able to update a Kubernetes Job, the Job has to be recreated by first being
+removed and the reapplied to the cluster.
 
-A typical use case for running Jobs with Flux is to implement post deployment tasks such as database scheme migration. This requires two seperate Kustomization resources to implement. One to create or recreate the Job and another to deploy the application.
+## Repository structure
 
-Given a simple Job in the path `./pre/job.yaml`.
+A typical use case for running Kubernetes Jobs with Flux is to implement pre-deployment tasks
+for e.g. database scheme migration and post-deployment jobs e.g. cache refresh.
+
+This requires separate [Flux Kustomization](../components/kustomize/kustomization.md) resources
+that depend on each other: one for running the pre-deployment Jobs,
+one to deploy the application, and a 3rd one for running the post-deployment Jobs.
+
+Example of an application config repository:
+
+```text
+├── pre-deploy
+│   └── migration.job.yaml
+├── deploy
+│   ├── deployment.yaml
+│   ├── ingress.yaml
+│   └── service.yaml
+├── post-deploy
+│   └── cache.job.yaml
+└── flux
+    ├── pre-deploy.yaml
+    ├── deploy.yaml
+    └── post-deploy.yaml
+```
+
+## Configure the deployment pipeline
+
+Given a Job in the path `./pre-deploy/migration.job.yaml`:
 
 ```yaml
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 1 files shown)


makkes:
https://deploy-preview-990--fluxcd.netlify.app/docs/use-cases/running-jobs/

content/en/docs/use-cases/running-jobs.md
```diff
@@ -0,0 +1,142 @@
+---
+title: Running pre and post-deployment jobs with Flux
+linkTitle: Running Kubernetes Jobs with Flux
+description: "How to run Kubernetes Jobs before and after an application deployment with Flux."
+weight: 10
+---
+
+Additional considerations have to be made when managing Kubernetes Jobs with Flux.
```
souleb, line 8: Maybe make that a `note`.

content/en/docs/use-cases/running-jobs.md
```diff
@@ -0,0 +1,142 @@
+---
+title: Running pre and post-deployment jobs with Flux
+linkTitle: Running Kubernetes Jobs with Flux
+description: "How to run Kubernetes Jobs before and after an application deployment with Flux."
+weight: 10
```
makkes, line 5: Not sure but I think this value determines the order of the page at https://deploy-preview-990--fluxcd.netlify.app/docs/use-cases/ and I'm not sure how the ordering is supposed to look like. @stefanprodan might know. Maybe we just don't care.

content/en/docs/use-cases/running-jobs.md
```diff
@@ -0,0 +1,142 @@
...
+---
+
+Additional considerations have to be made when managing Kubernetes Jobs with Flux.
+By default, if you were to have Flux reconcile a Job resource,
+it would apply it once to the cluster, the Job would create a Pod that can either error or run to completion.
+Attempting to update the Job manifest after it has been applied to the cluster is not be allowed, as changes to the
+Jobs `spec.Completions`, `spec.Selector` and `spec.Template` are not permitted by the Kubernetes API.
+To be able to update a Kubernetes Job, the Job has to be recreated by first being
+removed and the reapplied to the cluster.
```
makkes, line 14:
```suggestion
removed and then reapplied to the cluster.
```

content/en/docs/use-cases/running-jobs.md
```diff
@@ -0,0 +1,142 @@
...
+title: Running pre and post-deployment jobs with Flux
+linkTitle: Running Kubernetes Jobs with Flux
+description: "How to run Kubernetes Jobs before and after an application deployment with Flux."
+weight: 10
+---
+
+Additional considerations have to be made when managing Kubernetes Jobs with Flux.
+By default, if you were to have Flux reconcile a Job resource,
+it would apply it once to the cluster, the Job would create a Pod that can either error or run to completion.
+Attempting to update the Job manifest after it has been applied to the cluster is not be allowed, as changes to the
```
souleb, line 11:
```suggestion
Attempting to update the Job manifest after it has been applied to the cluster will not be allowed, as changes to the
```

content/en/docs/use-cases/running-jobs.md
```diff
@@ -0,0 +1,142 @@
...
+Jobs `spec.Completions`, `spec.Selector` and `spec.Template` are not permitted by the Kubernetes API.
```
souleb, line 12:
```suggestion
Job `spec.Completions`, `spec.Selector` and `spec.Template` are not permitted by the Kubernetes API.
```

content/en/docs/use-cases/running-jobs.md
```diff
@@ -0,0 +1,142 @@
...
+To be able to update a Kubernetes Job, the Job has to be recreated by first being
+removed and the reapplied to the cluster.
+
+## Repository structure
+
+A typical use case for running Kubernetes Jobs with Flux is to implement pre-deployment tasks
+for e.g. database scheme migration and post-deployment jobs e.g. cache refresh.
```
makkes, line 19:
```suggestion
for e.g. database scheme migration and post-deployment jobs (like cache refresh).
```
omits the repetition of "e.g." that's already used some words prior to this location.

content/en/docs/use-cases/running-jobs.md
```diff
@@ -0,0 +1,142 @@
...
+
+This requires separate [Flux Kustomization](../components/kustomize/kustomization.md) resources
+that depend on each other: one for running the pre-deployment Jobs,
+one to deploy the application, and a 3rd one for running the post-deployment Jobs.
+
+Example of an application config repository:
```
souleb, line 25:
```suggestion
Example of an application configuration repository:
```

makkes (commented):
Nice one! 🎉

content/en/docs/use-cases/running-jobs.md
```diff
@@ -0,0 +1,142 @@
...
+    name: my-app
+  path: "./deploy/"
+  interval: 60m
+  timeout: 5m
+  prune: true
+  wait: true
+```
+
+This means that the `app-deploy` Kustomization will wait until all the Jobs in `app-pre-deploy` run to completion.
+If the Jobs fail, the app changes will not be applied by the `app-deploy` Kustomization.
```
souleb, line 112:
```suggestion
If the Job fails, the app changes will not be applied by the `app-deploy` Kustomization.
```

souleb (commented):
maybe add a note about downtime

> In this model, in order to avoid down time, you have to make sure that your schema migration works on both versions of your application. 

Also we could add that the job encapsulate a db migration tool like `flyway` or `migrate`.

content/en/docs/use-cases/running-jobs.md
stefanprodan: I will fix the ordering and titles of all guides under use-cases in a a different PR. Thanks for bringing it up, I wanted to to this long time ago and I forgot about it.

stefanprodan:
> Also we could add that the job encapsulate a db migration tool like flyway or migrate.

I don't want to make this about flyway. If you watch Philip's talk about this, the use case is for running the migration from the app container with .NET entity framework. Having a flyway container breaks the current model, the whole idea is that you must bump all jobs and the app to the same image tag, otherwise the Jobs will not change and they will not run.

makkes (approved):

stefanprodan:
Thanks @makkes and @souleb for the review.

PR #992, opened 2022-06-10
Update Azure doc now that we support OCI Helm Repositories

souleb:
Signed-off-by: Soule BA <soule@weave.works>


stefanprodan (dismissed):
Have you tested it on Azure? Since we don't support auto-login, please add a a small guide for static auth with Docker config secrets and ACR.

souleb:
Sorry forgot about this. I added the requested guide.

hiddeco (approved):

PR #993, opened 2022-06-11
Fix concurrent description

lxm:
Fix concurrent description for notification-controller/helm-controller/source-controller


stefanprodan (approved):
LGTM 

Thanks @lxm

PR #994, opened 2022-06-12
Add --ssh-hostkey-algos to IAC docs

pjbgf:
Depends on https://github.com/fluxcd/image-automation-controller/pull/388


stefanprodan (changes requested):
Hold this until we release flux2 with IAC updated

stefanprodan (approved):

Issue #995, opened 2022-06-13
Code blocks small on mobile


dholbach:
![Image from iOS](https://user-images.githubusercontent.com/1346979/173309987-f0cdc593-5397-461b-b283-bc2af0bc4386.png)

Arhell:
/assign
url please

PR #1000, opened 2022-06-14
Update code block on mobile

Arhell:
resolve https://github.com/fluxcd/website/issues/995

### DIFF
--- a/assets/scss/_highlight.scss
+++ b/assets/scss/_highlight.scss
@@ -1,0 +1,5 @@
+.td-content .highlight {
+  @include media-breakpoint-down(md) {
+    max-width: 100%;
+  }
+}
--- a/assets/scss/main.scss
+++ b/assets/scss/main.scss
@@ -8,4 +8,5 @@
 @import "buttons";
 @import "terminal";
 @import "responsive-video";
+@import "highlight";
 @import "gallery";


Arhell:
cc @dholbach

dholbach:
Thanks @Arhell this is much better!

Do you think it'd make sense to make the overflow scrollbar visible in those case so it's easier to understand that there's content you can't see in that moment? Looks like Firefox and Chrome both don't show it for me in mobile view with more content in the codeblocks.

Arhell:
> Thanks @Arhell this is much better!
> 
> Do you think it'd make sense to make the overflow scrollbar visible in those case so it's easier to understand that there's content you can't see in that moment? Looks like Firefox and Chrome both don't show it for me in mobile view with more content in the codeblocks.

There is an attribute with a scrolling.
This is a feature of the browser on a mobile device.

I looked at examples on different sites, everything is the same there.

dholbach (approved):
Thanks for resolving this.

It looks the problem started appearing recently when https://github.com/google/docsy/blame/main/assets/scss/_code.scss#L6 behaviour was changed through https://github.com/google/docsy/commit/1731334ab7c402133eed28920336c35af04febbd#diff-1f4d7b8618cd04fd8c5bf1b1dc08c5ae99ade3e2418c79d1e8f2acaa07a90923R5

PR #1008, opened 2022-06-21
add missing repositories

dholbach:


hack/contributors.py
```diff
@@ -39,10 +39,13 @@ def get_contributions_from_gh():
     repos = [
         '.github',
         'community',
+        'charts',
```
stefanprodan, line 42: Only bots push changes to this one.

stefanprodan (approved):

PR #1011, opened 2022-06-23
GitHub Actions App Build guide

kingdonb:
When I was reviewing some of the docs I contributed earlier and trying to apply them to a project, I found the "Docker Build" example to be lacking in the existing "Manifest Generation" tutorial. It does what it needs to do for there, which is to briefly cover docker build, but only a side-note and an example of "what we told folks to do back when Flux v1 was widely used."

This doc shows a single GitHub Actions workflow that produces suitable builds for BOTH git tags (ideally tags which are valid semver numbers) AND git branches.

The old example of how to build a latest image only uses the git `SHA` or `REVISION` as a tag name and that's not suitable for Flux anymore (you should use Sortable Image Tags instead) or building images from semver tags. (The example is intentionally this way, it's meant to highlight that we can still use whatever 8-character SHA tagging strategy that Flux v1 users did, it just requires a bit of help from CI in the form of manifest generation.)

This example is supposed to be simpler, and follows all the guidance of Flux v2 from the ground up with no CI manifest generation. So this example does both (build from git tags AND git branches), so it should be a good example to highlight when we just want to show what is a minimum CI task needed to work with Flux that you should not outgrow.

I did not address caching at all, and I mentioned image provenance (Cosign) only to clarify that it is not covered here. I have assumed the Docker build is short enough that caching doesn't matter but a follow-on article might be worthwhile, showing more advanced CI based on this one, but with additions taking advantage of all Flux features.

### DIFF
--- a/content/en/docs/use-cases/gh-actions-app-builder.md
+++ b/content/en/docs/use-cases/gh-actions-app-builder.md
@@ -1,0 +1,248 @@
+---
+title: GitHub Actions Basic App Builder
+linkTitle: GitHub Actions Basic App Builder
+description: "How to build and push image tags for Flux from Git branches and tags."
+weight: 39
+---
+
+This guide shows how to configure GitHub Actions to build an image for each new commit pushed on a branch, for PRs, or for tags in the most basic way that Flux's automation can work with and making some considerations for both dev and production.
+
+A single GitHub Actions workflow is presented with a few variations but one simple theme: Flux's only firm requirement for integrating with CI is for the CI to build and push an image. So this document shows how to do just that.
+
+### Scope of this document
+
+Strictly speaking Flux considers CI to be out-of-scope, but this answer frequently leads to bad experiences caused by over-complicated CI built for users who did not firmly grasp the minimum requirements that Flux demands from a supporting CI. This example is intended to cover a majority of use cases with the simplest possible CI workflow.
+
+Users are not expected to strictly adopt this minimum viable solution or view this guidance as strongly prescriptive. You can adapt the example workflow for your use, and you can incorporate Flux's automation into your dev or production release machinery at a variety of critical points, mostly independent of one another.
+
+We anticipate in this guide that Flux users who are developing one or more apps likely want two build strategies for each app: a **Dev** build generates a (not semantically versioned) tag from some feature or environment branch with the branch name, commit hash, and timestamp; and a **Release** build produces a [semantic version] tag from the release tag that preceded it.
+
+You might want deployment automation in either or both environments, or perhaps neither. This guide shows how to generate image tags in a way that will be ready to work with Flux's automation for either or both of these scenarios.
+
+How to configure an `ImageUpdateAutomation` resource to take advantage of Release or Dev builds with automation is covered separately in the [Image Update Guide] and [Sortable image tags] guide, respectively.
+
+## Example GitHub Actions Workflow
+
+tl;dr: This build workflow does everything that Flux needs. Drop it into `.github/workflows/docker-build.yml` and reap the benefits.
+
+First copy this example and update `IMAGE` to point to your own image repository target. Then set `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` and you are done. Most git push events will now result in images suitable for Flux to deploy.
+
+For a deeper understanding and some variations, see the remainder of the doc.
+
+```yaml
+name: Docker Build, Push
+
+on:
+  push:
+    branches:
+      - '*'
+    tags-ignore:
+      - 'release/*'
+
+jobs:
+  docker:
+    env:
+      IMAGE: kingdonb/any_old_app
+    runs-on: ubuntu-latest
+    steps:
+      - name: Prepare
+        id: prep
+        run: |
+          BRANCH=${GITHUB_REF##*/}
+          TS=$(date +%s)
+          REVISION=${GITHUB_SHA::8}
+          BUILD_ID="${BRANCH}-${REVISION}-${TS}"
+          LATEST_ID=canary
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            BUILD_ID=${GITHUB_REF/refs\/tags\//}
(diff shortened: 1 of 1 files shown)


kingdonb:
Side-note: I did not yet consider the placement of the article in the TOC list eg. `weight: 50` so I have marked this as a draft. I'll make an adjustment but my perception right now is this guide is basically ready for review, other than that one detail.

I wanted to get a look in the Preview before marking it as ready for review, but I am out of time for this today, anyone is welcome to submit feedback now.

content/en/docs/use-cases/gh-actions-app-promotion.md
```diff
@@ -0,0 +1,205 @@
+---
+title: GitHub Actions Basic App Promotion
```
stefanprodan, line 2: For this title I see very different guide where we use `flux build artifact` and `flux tag artifact` to showcase a promotion workflow for the whole app definition (container+manifests).

Could we use "GitHub Actions Basic App Builder" as the title instead?

kingdonb (commented):
Change the introduction

content/en/docs/use-cases/gh-actions-app-builder.md
```diff
@@ -0,0 +1,219 @@
...
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
```
kingdonb, line 76:
```suggestion
        uses: docker/build-push-action@v3
```

kingdonb: This version is out of date

kingdonb:
In the linked PR, I found that caching is not that hard to add, but there are a few things I tripped over. I almost think it's worth it to address caching.

```
- name: Build and push
  id: docker_build
  uses: docker/build-push-action@v3
  with:
    context: .
    push: true
    tags: |
      ${{ env.IMAGE }}:${{ steps.prep.outputs.BUILD_ID }}
      ${{ env.IMAGE }}:${{ steps.prep.outputs.LATEST_ID }}
    cache-from: type=registry,ref=ghcr.io/kingdonb/river-data-explorer:buildcache
    cache-to: type=registry,ref=ghcr.io/kingdonb/river-data-explorer:buildcache,mode=max
```

Then again I'm not sure – we would have to make a choice, for example I chose here the registry cache mode, because I used a multi-stage build and it took me a while to find this doc: [build-push-action/cache.md](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md) which explains about cache modes min and max, and why you might need to choose the registry mode instead.

Maybe we can leave the details about caching out, but just add a link to this document and suggest it as a follow-up reading. It has a lot of helpful information for example, did you know that `context: .` is required, else caching will not work? It's not mentioned in this document, except that all the examples just do it. I can't tell you how many times I've tried to implement caching docker builds on GitHub Actions and I never saw that, (now it's no wonder it never worked for me before! 🤯 )

stefanprodan:
@kingdonb maybe we should publish this as a blog post? Docker actions do change a lot, the caching was broken some time ago, so we had to do all sorts of tricks in Flux, see https://github.com/fluxcd/notification-controller/blob/main/.github/workflows/e2e.yaml#L72. There is a bug chance the examples in this doc will not work in a month or two. If you don't plan to maintain this on long run, better to have this as a blog post.

kingdonb:
That's one reason why I thought it would be a good idea to avoid a discussion of caching.

Maybe there should be a blog post instead, but whatever is in the article right now, I think is pretty ever-green. I'd be in favor of merging it here as-is, then maybe publish a blog post about it later.

content/en/docs/use-cases/gh-actions-app-builder.md
```diff
@@ -0,0 +1,219 @@
+---
+title: GitHub Actions Basic App Builder
+linkTitle: GitHub Actions Basic App Builder
+description: "How to build and push image tags for Flux from Git branches and tags."
+weight: 39
+---
+
+This guide shows how to configure GitHub Actions to build an image for each new commit pushed on a branch, for PRs, or for tags in the most basic way that Flux's automation can work with and making some considerations for both dev and production.
+
+In the [GitHub Actions Manifest Generation] guide, we saw some workflows that render manifests in CI and commit them back to Git. From [Image Update Guide] we saw that Flux can set `.spec.git.push.branch` to [Push updates to a different branch] than the one used for checkout. Those advanced techniques are above and beyond what is needed to work with Flux.
```
stefanprodan, line 10: If we send people from here to a guide that is 10x the size of this, we'll lose them. I think we should remove this from here, and put it down below at other links.

kingdonb: I agree, I have made this change in [428cdad](https://github.com/fluxcd/website/pull/1011/commits/428cdadebdfbe64207521ed734a384e0c61f7310) – manifest generation is still mentioned, but less prominently and at the end.

The reference is now listed as part of a "Further Reading" section at the bottom that also provides some closure about how the reader should follow the Flux docs to make use of these built images from CI.

I added another sub-heading for caching, it provides no specific guidance other than fundamental "fast builds are better than slow ones" and only references other documentation to provide specifics so should hopefully remain evergreen, even if there are changes in the ecosystem.

kingdonb:
I'm giving it another read through at this point to check for errors, then I will squash and write back here when I think it's ready for one last review. 👍

kingdonb:
[ef43017](https://github.com/fluxcd/website/pull/1011/commits/ef43017bb4c88ee86b0a2f81477f07f0bdde32cd) is the last commit before squash, rebasing now. I've reordered the "extra resources" at the end and added some additional text to help everything flow better.

stefanprodan (approved):
LGTM 

Thanks @kingdonb

PR #1017, opened 2022-06-29
Add SenseLabs to adopters

netthier:
We are using Flux for both internal development clusters as well as production clusters.


stefanprodan (approved):
Thanks @netthier 🥇

PR #1023, opened 2022-07-05
Redesign ecosystem page

Arhell:


dholbach:
Did you try using `tabpane` and `tab` that are included in recent `docsy` setups?

I'm not quite sure how we'd convey the same level of information as we currently do, but one idea was to make it look the page a little more like https://istio.io/latest/about/ecosystem/ and show some more logos.

dholbach:
Maybe for now it could be better to start thinking about a way how to nicely show logos as well. The tab idea is maybe less important.

Should we do the text in a hover-over box? Is that accessible? Or should we do a table like this:

`| logo | name / link  | description |`

Any opinions from anyone?

Arhell:
I can do like in istio, the problem is that some logos are repeated, for example weaveworks and microsoft

dholbach:
Maybe we don't need tabs necessarily, but another nice way of laying out the different parts of our ecosystem ... even if there's some repetition. 

I asked other folks for input too.

staceypotter (changes requested):
Thanks for this, it looks pretty good! I know you're still in draft mode, but I did notice there's no padding between the lines for these tabs and just wanted to make a note of that now. :)

dholbach (approved):
Wow, well done @Arhell !

PR #1029, opened 2022-07-08
play around - trying to change ecosystem page to behave similarly as …

dholbach:
…adopters

### DIFF
--- a/assets/scss/_ecosystem.scss
+++ b/assets/scss/_ecosystem.scss
@@ -1,0 +1,42 @@
+/* ecosystem page */
+.ecosystem {
+    .card-deck {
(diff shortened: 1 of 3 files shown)


dholbach:
If anybody wants to take a look at styling https://deploy-preview-1029--fluxcd.netlify.app/ecosystem/ or how to differently present the content, please go ahead - that'd be great. In the meantime I'd take a look at auto-generating this from YAML.

dholbach:
@stefanprodan says
> I would pick 3 vendors to show their logs only, then use tables everywhere, the 3 vendors should show up in the table as well
> For the UI section we need print screens then the table (bearbeitet) 
> for all the other sections, just tables

PR #1036, opened 2022-07-12
Update packages.md

chengleqi:
Is it a typo? cause `5 + time.Minute` means 1m0.000000005s.


stefanprodan (approved):
LGTM

Thanks for the fix @chengleqi

PR #1037, opened 2022-07-12
Add Mintmesh to adopters

kirankldevops:
Signed-off-by: kirankldevops <kiran.kl@mintmesh.com>


stefanprodan (approved):
Thanks @kirankldevops 🏅

PR #1050, opened 2022-07-21
Fix Secret Decryption

pjbgf:
The only controller that supports auto decryption is Kustomize. This removes the inaccurate reference to Helm on this context.

An additional commit resolves presentation issues due to the `details` block with was making the Markdown to be interpreted as HTML. Here's the before/after:
image

### DIFF
--- a/content/en/docs/security/best-practices.md
+++ b/content/en/docs/security/best-practices.md
@@ -121,7 +121,7 @@ The recommendations below are based on Flux's latest version.
 - Ensure Secret Decryption is enabled and secrets are not being held in Flux Sources in plaintext.
   <details>
     <summary>Rationale</summary>
-    Helm and Kustomize Controllers have auto decryption mechanisms that can decrypt cipher texts on-demand at reconciliation time. This enables credentials (e.g. passwords, tokens) and sensitive information to be kept in an encrypted state in the sources.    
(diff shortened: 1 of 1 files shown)


content/en/docs/security/best-practices.md
```diff
@@ -112,7 +112,7 @@ The recommendations below are based on Flux latest version.
 - Ensure Secret Decryption is enabled and secrets are not being held in Flux Sources in plain-text.
   <details>
     <summary>Rationale</summary>
-    Helm and Kustomize Controllers have auto decryption mechanisms that can decrypt cipher texts on-demand at reconciliation time. That enables for credentials (e.g. passwords, tokens) and sensitive information to be kept encrypted in the sources.    
+    The Kustomize Controller have an auto decryption mechanism that can decrypt cipher texts on-demand at reconciliation time. That enables for credentials (e.g. passwords, tokens) and sensitive information to be kept encrypted in the sources.
```
hiddeco, line 115: Should we also make mention of it making use of SOPS? Something like

> [..] using an embedded implementation of [SOPS].

pjbgf: Added the complement accordingly, PTAL.

pjbgf:
Preview page can be seen at: https://deploy-preview-1050--fluxcd.netlify.app/docs/security/best-practices/

hiddeco (approved):

Issue #898, opened 2022-04-25
Ecosystem page: show the tooling flux integrates with


dholbach:
It'd be good to explain very prominently how easily and deeply Flux integrates with Helm, Kustomize, Vault and other tools.

PR #1054, opened 2022-07-25
ecosystem: show best-practice OSS we integrate with

dholbach:
Fixes: #898


dholbach:
Please let me know if there are obvious entries missing.

And please send help styling this nicely. :-D

dholbach:
→ https://deploy-preview-1054--fluxcd.netlify.app/ecosystem/

content/en/ecosystem/index.md
```diff
@@ -16,6 +16,45 @@ If you represent an organisation, please add yourself to [the adopters page](/ad
...
+{{% /card %}}
+{{% card header="[Kubernetes](https://kubernetes.io/)"
+         footer="[Docs](/docs/get-started/)" %}}
+![Kubernetes](./img/kubernetes.svg)
+{{% /card %}}
+{{% card header="[Kyverno](https://kyverno.io/)"
+         footer="[Docs](/blog/2022/02/security-image-provenance/)" %}}
+![Kyverno](./img/kyverno.png)
+{{% /card %}}
+{{% card header="[Prometheus](https://kubernetes.io/)"
```
pjbgf, line 45:
```suggestion
{{% card header="[Prometheus](https://prometheus.io/)"
```

pjbgf (approved):
Although it is marked as draft, it is LGTM for me. 

Reviewing as per request during Flux Dev Meeting.

makkes (approved):
I love it!

dholbach:
@stefanprodan Do you have any thoughts?

stefanprodan:
I don't like the bottom border for the docs link, this also take a lot of space pushing all the other section outside of the view. Can we move the doc links on the logos?

stefanprodan (commented):
I'm on the fence with this, how did we end up chasing these tools out of tens of tools that Flux and Flagger integrates with. This will probably result in people being angry their projects are not listed here. If we put Istio there, then why not all the other 10+ solutions Flagger works with? I think we need a dedicated page for this where we list all the tools.

dholbach:
> I'm on the fence with this, how did we end up chasing these tools out of tens of tools that Flux and Flagger integrates with. This will probably result in people being angry their projects are not listed here. If we put Istio there, then why not all the other 10+ solutions Flagger works with? I think we need a dedicated page for this where we list all the tools.

A separate page is a good idea.

dholbach:
I've been trying to come up with a good place for a separate page and am considering putting a section (with smaller logos?) to https://fluxcd.io/docs/ and maybe drop the "flux in short" or move it down?

It'd be good to 
1. show that we integrate well with a wealth of different tooling
2. point people to docs or examples of how to get the most out of the integration

`/docs` might be a good place for this... WDYT?

PR #1063, opened 2022-08-02
Remove duplicate AWS EKS Anywhere and Weave GitOps entries

dholbach:
- AWS EKS Anywhere was already listed under Product/Services
- Weave GitOps was already listed under UIs


dholbach:
I was considering moving `weaveworks/vscode-gitops-tools` under UIs as well ... WDYT?

stefanprodan (approved):

dholbach:
@kingdonb what do you think about moving the Vs code entry?

kingdonb:
@dholbach Yes I agree, "Extensions" is technically correct but in a slightly upside-down sense, because this is not a Flux extension it's a VSCode extension for Flux.

If we had a new section or page for Flux UIs then the VS Code Extension should go there. 👍

dholbach:
We do... https://fluxcd.io/ecosystem/#flux-uis :-)

kingdonb:
Got it! I'll send a PR

dholbach:
I can do it do - no problemo.

kingdonb:
Here it is: https://github.com/fluxcd/website/pull/1064

PR #1076, opened 2022-08-11
Add missing `apiVersion` fields to examples in the Repository Structure page

fr3fou:
Currently, the examples in the [Ways of structuring your repositories](https://fluxcd.io/docs/guides/repository-structure/) section are broken. 
The Kubernetes manifests are missing an `apiVersion` property which means they'll fail when applying.
This PR adds the missing properties to every example on that page.

### DIFF
--- a/content/en/flux/guides/repository-structure.md
+++ b/content/en/flux/guides/repository-structure.md
@@ -155,6 +155,7 @@ App repository plain Kubernetes manifests example:
 Delivery example (stored in config repo):
 
 ```yaml
+apiVersion: source.toolkit.fluxcd.io/v1beta2
 kind: GitRepository
 metadata:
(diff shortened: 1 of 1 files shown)


kingdonb:
This looks like a reasonable change to me, except that it's another place we must update in the examples whenever the API version has changed, and I don't know if we already have a plan in place to deal with those. We must grep for them. It should definitely not block from merging this fix, since there's no place those YAMLs are valid without an apiVersion.

Maybe better to give a quick grep through for any outdated apiVersion references through the rest of the docs before we close this. I'll approve it now, at least, as it's for sure technically sound.

fr3fou:
Just rebased the branch with upstream, hopefully we can merge this ^^

PR #1077, opened 2022-08-14
Fix link to ConfigMap generator documentation

Benny-Git:
Signed-off-by: Benny <benjamin@bertow.com>


dholbach (approved):
Thanks a lot for the fix!

PR #1084, opened 2022-08-20
docs: Update flux-v1-migration.md

suryapandian:
Signed-off-by: Surya <surya.pandian@truelayer.com>


kingdonb:
Thanks for making this PR! I'm thinking about changing the wording to be a bit more positive. What you wrote is basically factual but from the perspective of a Flux v1 user, it's not all that has changed. When Flux v1 failed to apply a commit, it would simply emit errors in the log and refuse to advance the flux-sync tag. It was up to users or monitoring to notice that something has gone wrong, as subsequent commits could be partially applied as well, leaving certain changes behind with a syntax error that has gone unnoticed.

I'd like to explain what has changed for the user in a slightly more positive way, while also highlighting these things they need to know (like that you are supposed to consider breaking your Flux workloads apart into units or Flux Kustomizations), and that the point of doing this is to isolate failures and ensure that those units can be applied either completely or not at all.

Please give me a bit to consider some alternate wording and get back to you! Thanks again for this contribution and 🎉 hello First-Time Contributor, big welcome from the Flux community :)

content/en/docs/migration/flux-v1-migration.md
```diff
@@ -19,6 +19,12 @@ design.
 This may at times mean that you have to make adjustments to the way your
 current cluster configuration is structured. If you are in this situation
 and need help, please refer to the [support page](/support/).
+
+One of the interesting features of flux2 is the "server-side validations" of the manifests. 
+It is important to note that the manifests are applied to the cluster only if the validation passes.
+If one manifest has multiple resources, an error in one of the resources would also prevent other resources from getting applied. 
+This was not the behavior with flux1
```
kingdonb, line 26:
```suggestion
One key change in Flux v2 is "server-side apply" that enforces strict
validation of the manifests.  It is important to note that manifests are
applied atomically to the cluster only if YAML validation passes. If one
Kustomization has multiple resources, an error in any one of the resources
will also prevent other resources in that group from getting applied. This is
a breaking change from Flux v1.
```

I think this conveys the advantage of doing things this way without papering over, the words "breaking change" probably belong in whatever version of this we accept 👍 

Thanks for suggesting this update

kingdonb (commented):
I have suggested a new message, please let me know if this gets the right message across @suryapandian 🤔

content/en/docs/migration/flux-v1-migration.md
```diff
@@ -19,6 +19,14 @@ design.
 This may at times mean that you have to make adjustments to the way your
 current cluster configuration is structured. If you are in this situation
 and need help, please refer to the [support page](/support/).
+
+One key change in Flux v2 is "server-side apply" that enforces strict
+validation of the manifests.  It is important to note that manifests are
+applied atomically to the cluster only if YAML validation passes. If one
```
kingdonb, line 25: "only if YAML validation passes" seems a bit technically imprecise but I think this is the right amount of detail for this early in the guide. Should we try to mention this again later, is there a right time in the progression of the text? WDYT

kingdonb:
Looks like this was accidentally closed when you rebased, or has been handled in another branch. (Did you mean to close this one?)

dholbach:
This change seems to toll back a lot of changes. Can you try to summarise what this PR is trying to do?

suryapandian:
@kingdonb @dholbach While trying to rebase ended up deleting the forked repository and this PR became messy. COuld you please review https://github.com/fluxcd/website/pull/1286 the changes are moved here. Thank you 🙇

PR #1088, opened 2022-08-23
docs: Add secrets management page

pjbgf:


content/en/docs/security/secrets-management.md
```diff
@@ -0,0 +1,276 @@
...
+
+This is transparent to the consuming applications, making it a quite suitable approach to retrofit
+into an existing setup. An example of a Secret Decryption Operator is [Sealed Secrets]. 
+
+Storing encrypted secrets in Git repositories enables configuration versioning to leverage the
+same practices used for managing, versioning and releasing application and/or infrastructure
+when using declarative "everything as code", for example pull-requests, tags and branch
+strategies.
+
+Note that some sources may keep a history of the encrypted Secrets (e.g. `GitRepository`)
```
aryan9600, line 58: maybe it's worth mentioning that a source containing a secret (even if encrypted) should always be private? raising this because of the following line in https://fluxcd.io/docs/guides/sealed-secrets/
> In order to store secrets safely in a public or private Git repository, you can use Bitnami’s [sealed-secrets controller](https://github.com/bitnami-labs/sealed-secrets)

same goes for the SOPS section as well.

pjbgf: I am not sure we want to be that prescriptive, as each use case has different security requirements. For example, this could be a valid use amongst public repositories that share non critical secrets for example.

I tried to cover the high-level concern in:
- Access Management and Auditing: "having secrets stored (even if in encrypted form) in an easily accessible Flux source that has a loosely set of RBAC and no auditing in place may not meet such requirements."
- Summary: "ensure the appropriate level of security controls are in place"

content/en/docs/security/secrets-management.md
```diff
@@ -0,0 +1,276 @@
...
+
+Access to secrets should be restricted to the servers and applications within the environment
+they need to be accessed. The same goes to users and service accounts.
+
+When considering the different solutions, it is important to note how they hang together
+and what the gaps are. If you have a strong requirement for access and auditing controls,
+having a well-defined api-server auditing in place, together with tight RBAC policies in your
+cluster is only part of the problem. Also take into account how those secrets are sourced,
+stored and handled. Maybe having secrets stored (even if in encrypted form) in an easily
+accessible Flux source that has a loosely set of RBAC and no auditing in place may not meet
```
fire-ant, line 186: nit "loosely defined RBAC"

content/en/docs/security/secrets-management.md
```diff
@@ -0,0 +1,276 @@
...
+and have a short expiration time. In some cases you can even tie them to a network boundary, 
+meaning that even if they get breached, they won't be able to be used outside the current
+context. 
+
+Flux supports [contextual authorization] for the major Cloud Providers, be aware of the supported
+features and use them whenever possible.
+
+### Detect "chicken and egg" scenarios
+
+Flux won't protect yourself from yourself. On a running cluster, it is quite easy to incrementally
```
fire-ant, line 247: "won't protect you from yourself" this maybe reads more naturally (though both are acceptable!)

stefanprodan (approved):
LGTM

Awesome job, thanks @pjbgf 🥇

Issue #1089, opened 2022-08-23
docs navigation on hover bug


Arhell:
/assign
![Screenshot_1](https://user-images.githubusercontent.com/26163841/186211640-3a01ed01-2b7c-484a-917e-84025dda444f.png)

PR #1090, opened 2022-08-23
fix docs navigation on hover

Arhell:
resolve https://github.com/fluxcd/website/issues/1089


dholbach:
I wonder if we could send this to `google/docsy` as a PR? What do you think?

Arhell:
> I wonder if we could send this to `google/docsy` as a PR? What do you think?

Maybe when there is time.

Arhell:
Close via https://github.com/google/docsy/issues/1198

PR #1091, opened 2022-08-24
doc: missing parameter in example

sympatheticmoose:
Missing `spec.prune` causes example to fail to reconcile:

```
Kustomization reconciliation failed: Kustomization/oci/podinfo dry-run failed, reason: Invalid, error: Kustomization.kustomize.toolkit.fluxcd.io "podinfo" is invalid: spec.prune: Required value
```

Signed-off-by: David Harris <david.harris@weave.works>


stefanprodan (approved):
LGTM

Thanks @sympatheticmoose

PR #1092, opened 2022-08-24
Clarify how to make rollout restarts work with flux-client-side-apply

annaken:


content/en/docs/faq.md
```diff
@@ -191,12 +191,20 @@ In order for Flux to preserve fields added with kubectl, for example a label or
 you have to specify a field manager named `flux-client-side-apply` e.g.:
 
 ```yaml
-kubectl annotate --field-manager=flux-client-side-apply 
+kubectl --field-manager=flux-client-side-apply label ...
```
makkes, line 194: Why `label` instead of `annotate`?

annaken: The fact that the rollout restart is failing on an additional annotation made this line super confusing to me - so I just wanted to be clear that this is a generic example and not the instruction about how to edit that restartedAt annotation.

makkes: I understand, that sounds reasonable. How about we change the sentence right ahead of this example to something like "e.g. for manually adding a label to a resource..."?

makkes (approved):
Great job, thanks!

stefanprodan:
@annaken can you please [signoff](https://github.com/fluxcd/website/pull/1092/checks?check_run_id=8011591052) your commit and force push it? Thanks

PR #1099, opened 2022-08-26
Add example for enabling notifications for 3rd party controllers

stefanprodan:
Add bootstrap patch example for enabling notifications for third party controllers.

### DIFF
--- a/content/en/docs/cheatsheets/bootstrap.md
+++ b/content/en/docs/cheatsheets/bootstrap.md
@@ -445,3 +445,40 @@ patches:
       kind: Kustomization
       name: "flux-system"
 ```
+
+### Enable notifications for third party controllers
+
+Enable notifications for 3rd party Flux controllers such as [tf-controller](https://github.com/weaveworks/tf-controller):
+
+```yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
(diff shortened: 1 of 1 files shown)


hiddeco (approved):

content/en/docs/cheatsheets/bootstrap.md
```diff
@@ -445,3 +445,40 @@ patches:
...
+  - patch: |
+      - op: add
+        path: /rules/-
+        value:
+          apiGroups: [ 'infra.contrib.fluxcd.io' ]
+          resources: [ '*' ]
+          verbs: [ '*' ]
+    target:
+      kind: ClusterRole
+      name:  crd-controller
```
chanwit, line 483: There's no `crd-controller` inside my `gotk-components.yaml`, but there is `crd-controller-flux-system`.
I guess there was a mechanism that added the namespace as a suffix to the ClusterRole, right?

chanwit (changes requested):

content/en/docs/cheatsheets/bootstrap.md
stefanprodan: Yes good catch, I’ll fix it. Thanks

PR #1118, opened 2022-08-31
Add detail about Kustomize controller timeout

kingdonb:
The Kustomize Controller's reconciling can take up to 3x `spec.timeout`

(This was an action item from today's Flux Dev meeting)

It seemed to fit naturally in this short paragraph about the Kustomize Server-Side Aply timeout that was missing details. Is there anywhere else we can highlight this, perhaps in the Kustomize controller docs as well


stefanprodan (approved):
Thanks @kingdonb

PR #1130, opened 2022-09-05
Add ssh_config warning in installation.md

snebel29:
This is a follow-up change from: https://github.com/fluxcd/flux2/issues/3045


stefanprodan (approved):
LGTM

Thanks @snebel29 🏅

PR #1138, opened 2022-09-08
Add "FAQ" text to the description for FAQ page

### LINKED ISSUE #1137: Can't find Flux FAQ from fluxcd.io easily
Hi! I was on fluxcd.io and had heard of Flux FAQs. I couldn't find them until I was digging around in the docs. Is there a way that we can make it easier to find from fluxcd.io ? I searched "faq: in the fluxcd.io search bar and it only showed me the flagger faq link (which is good). I hope my question makes sense. thanks!

kingdonb:
Fix #1137

Not certain if this will help (the main FAQ page is not ranked at all when you search for "FAQ")

### DIFF
--- a/content/en/flux/faq.md
+++ b/content/en/flux/faq.md
@@ -1,7 +1,7 @@
 ---
 title: "Frequently asked questions"
 linkTitle: "FAQ"
-description: "Flux and the GitOps Toolkit frequently asked questions."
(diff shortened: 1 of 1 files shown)


kingdonb:
(Can we test changes to the search, or does it only index from the live site?)

kingdonb:
If this change will have any effect, it doesn't appear to be possible to tell from the preview site.

PR #1140, opened 2022-09-09
Recommend OCI as a Bucket alternative rather than chartmuseum

makkes:
Signed-off-by: Max Jonas Werner <max@e13.dev>

### DIFF
--- a/content/en/flux/guides/helmreleases.md
+++ b/content/en/flux/guides/helmreleases.md
@@ -204,71 +204,7 @@ as the whole storage bucket will be downloaded by source controller at each sync
 bucket can easily become very large if there are frequent releases of multiple charts
 that are stored in the same bucket.
 
-A better option is to use [Chartmuseum](https://github.com/helm/chartmuseum) and run a cluster
-local Helm repository that can be used by source controller. Chartmuseum has support
-for multiple different cloud storage solutions such as S3, GCS, and Azure Blob Storage,
-meaning that you are not limited to only using storage providers that support the S3 protocol.
-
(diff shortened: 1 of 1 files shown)


kingdonb:
Thanks! Do you think we should move the ChartMuseum doc somewhere else?

People can always look it up in git history, but it's basically rendered inaccessible to ordinary users this way. (On the other hand, the doc gets shorter and we are full-steam ahead on OCI...)

stefanprodan (approved):

kingdonb (approved):

kingdonb:
(Here is a blog post for the attic, in order to close the loop:
https://blog.teamhephy.info/blog/posts/attic/chartmuseum-and-what-comes-after.html)

The content we deleted from the docs is there, preserved for some posterity. We talked about maybe adding an official Flux attic but I think this is fine for now.

jimdickinson:
This kind of content (self host chartmuseum in-cluster) is what made me want to use Flux fwiw. Thanks @kingdonb for preserving it, but it took me an hour to track this down and I knew I'd read it before. I don't think users new to Flux will be able to find it.

kingdonb:
You are not the only person who has been looking for guidance on S3 and how to host charts there since we merged this PR, @jimdickinson. It's a shame we didn't wipe out a permalink/anchor so that we could put it back now – anyway, I have had questions about how to use a public S3 bucket and a private S3 bucket as a Helm Repository since this got merged, and some of those people are not ready to adopt OCI charts yet.

Perhaps we should consider the Flux attic idea, I'm not sure how else we can preserve this useful doc without making each purpose-focused guide longer and unwieldy with too many options and too little opinion about which one you should pick!

PR #1156, opened 2022-09-20
Update timetable and set Flux v1 sunset for November 1st

stefanprodan:
Given that Flux v1 and helm-operator dependencies are pinned to EOL versions (Kubernetes 1.21, Kustomize 3.8), which cannot be upgraded without causing regressions or a cascading amount of changes to the codebase, I propose we archive both projects on 1st Nov 2022, right after Kubernetes v1.22 reaches end-of-life on 28th Oct 2022.

Migration timetable updates:
- Add all supported API versions
- Add helm-operator to the archival list
- Set November 1st 2022 as the date for when Flux v1 and helm-operator repositories will be archived

Refs:
- Flux v1 archival https://github.com/fluxcd/flux/issues/3645
- Helm Operator archival https://github.com/fluxcd/helm-operator/issues/667


darkowlzz (approved):

kingdonb (approved):
👍 from me for this

hiddeco (approved):

phillebaba (approved):
Plenty of time has been given to move to Fluxv2. I know of some of the issues making Helm Operator work with newer k8s versions, and its not worth the time.

pjbgf (approved):

PR #1160, opened 2022-09-22
changed "clusters" to "cluster" under Add podinfo repository to Flux

syedabrarali:
changed "clusters" to "cluster" under Add podinfo repository to Flux because when the repo gets created it is ./cluster/my-cluster/flux-system but in the document it is clusters/my-cluster/flux-system

Signed-off-by: syedabrarali <50911126+syedabrarali@users.noreply.github.com>


stefanprodan:
It should be `clusters` everywhere in the doc.

PR #1165, opened 2022-09-27
HelmRelease Guide: Values Secret generated with Kustomize and SOPS

kingdonb:
One more additional guide for HelmReleases with values that come from a SOPS encrypted secret.

There are a few things to be aware of that make this not quite straightforward extrapolation from the ConfigMapGenerator example, I think this should be good to help more users from struggling. This has come up several times in Slack, we finally had a user who was persistent enough to discover all of the major tripping hazards and help me document them here.

We may want to let this one gather some reviews before it merges, I have tested it before writing, so strictly speaking this doc is not yet "tested" as in shown to work from a reading of the document, but the methods are tested and verified to work.

Thanks to our Slack community for helping us cover all our bases in the docs! 🌮 🎉

### DIFF
--- a/content/en/flux/guides/helmreleases.md
+++ b/content/en/flux/guides/helmreleases.md
@@ -385,6 +385,127 @@ removed from the cluster by kustomize-controller if
 [pruning](../components/kustomize/kustomization/#garbage-collection) is enabled.
 {{% /alert %}}
 
+## Refer to values in Secret generated with Kustomize and SOPS
+
+It is possible to use Kustomize [Secret generator](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/secretgenerator/)
+to trigger a Helm release upgrade every time the encrypted secret values change.
+
+A SOPS configuration for your cluster is required first. Follow the [Manage Kubernetes secrets with Mozilla SOPS](../mozilla-sops/)
+guide. The details of configuring SOPS are out of scope for this entry.
+
+Once you have SOPS configured, create a `kustomizeconfig.yaml` for Kustomize to
+be able to patch `Secrets` referenced in `HelmRelease` manifests:
+
+```yaml
+nameReference:
+- kind: Secret
+  version: v1
+  fieldSpecs:
+  - path: spec/valuesFrom/name
+    kind: HelmRelease
+```
+
+Create a `HelmRelease` definition that references a `Secret`:
+
+```yaml
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: podinfo
+  namespace: podinfo
+spec:
+  interval: 5m
+  releaseName: podinfo
+  chart:
+    spec:
+      chart: podinfo
+      sourceRef:
+        kind: HelmRepository
+        name: podinfo
+  valuesFrom:
+    - kind: Secret
+      name: podinfo-values
+```
+
+Ensure that this HelmRelease will be applied from the same Flux Kustomization
+that will decrypt the values secret. The encrypted secret data must be housed
+in the same Flux Kustomization path as the `HelmRelease` in order to allow
+Kustomize's Secret generator function to compose them together with the
+Kustomize configuration.
+
+Create a `kustomization.yaml` that generates the `Secret`:
+
+```yaml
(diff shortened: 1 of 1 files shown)


kingdonb:
I spent a couple of extra minutes to figure out what kind of `creation_rules` in `.sops.yaml` might permit users to encrypt a `values.yaml` file which is not a well-formed Kubernetes secret, and came up with something that I've now added in a subsequent commit.

kingdonb:
Credit goes to @jseiser who worked out how to do this mostly on his own, with a little bit of help in our Slack channel 🎉 

https://gist.github.com/jseiser/eed3b4ffdd1ff3def604c4c78c989c1f
^ his original version is a bit more concise

kingdonb:
Squashing for review

pjbgf (approved):
LGTM

Great job @kingdonb! :bow:

content/en/flux/guides/helmreleases.md
```diff
@@ -385,6 +385,133 @@ removed from the cluster by kustomize-controller if
...
+creation_rules:
+  - path_regex: .*values.yaml$
+    pgp: KEY_ID_ASDF1234
+  - path_regex: ".*\.yaml"
+    encrypted_regex: ^(data|stringData)$
+    pgp: KEY_ID_ASDF1234
+```
+
+Alternatively we could rename the input file to avoid matching, or handle the
+encryption elsewhere, like in CI where it can be made safer to handle secrets.
```
pjbgf, line 501: After talking with @stefanprodan, I think we could remove this opens a whole can of worms around the security of the entire process which is beyond the scope here.

```suggestion
```

content/en/flux/guides/helmreleases.md
```diff
@@ -385,6 +385,133 @@ removed from the cluster by kustomize-controller if
...
+Run this command to encrypt to the output file that was named in
+`kustomization.yaml` above:
+```
+$ sops -e --input-type=yaml --output-type=yaml my-values.tmp.secret > my-values.enc.yaml
+```
+
+Commit the `my-values.enc.yaml` file and discard the temp file, being sure not
+to accidentally commit secrets to the repository. Read on for more information.
+
+If users have followed the SOPS guide then we likely added a `creation_rules`
```
pjbgf, line 471:
```suggestion
If users have followed the [SOPS guide](../mozilla-sops/) then we likely added a `creation_rules`
```

content/en/flux/guides/helmreleases.md
```diff
@@ -385,6 +385,131 @@ removed from the cluster by kustomize-controller if
...
+Commit the `my-values.enc.yaml` file and discard the temp file, being sure not
+to accidentally commit secrets to the repository. Read on for more information.
+
+If users have followed the [SOPS guide](../mozilla-sops/) then we likely added a `creation_rules`
+entry telling the `sops` CLI how to handle encrypting all YAML secrets, like:
+
+```yaml
+# ./apps/sensitive/.sops.yaml
+creation_rules:
+  - path_regex: ".*\.yaml"
```
stefanprodan, line 477: This rule does not match `my-values.tmp.secret` so how come is an issue?

kingdonb: Some explanation is definitely necessary because the failure mode is very bad. If people take `values.yaml` and try to encrypt it directly, it will come out completely unencrypted with the configuration that is put forward in the SOPS guide, unless it had a section `data` or `stringData`. Plus I really didn't want to write "don't call your values.yaml file `values.yaml`" it just sounds wrong. I really wanted to include stronger language and a more specific method that doesn't add any risk the unencrypted file will accidentally get committed before it gets encrypted.

I actually wanted to mention in a separate document something like the unencrypted secret precommit hook that prevents you from committing and pushing a secret that hasn't been encrypted, I stumbled upon the other day. But it necessarily depends on the secrets being well-formed Kubernetes secrets to detect and block them from being committed. I have no mitigation strategy to prevent folks from accidentally committing a values.yaml that hasn't been encrypted properly, except to warn people against mixing secrets and non-secrets, and advising directly to use some caution when handling secrets.

stefanprodan: Let's use `values.yaml` please

stefanprodan (approved):
LGTM

Thanks @kingdonb

PR #1167, opened 2022-09-29
Add Cosign signing and verification to OCI docs

stefanprodan:
Add Cosign signing and Flux verification to the OCI artifacts guide.

Part of: https://github.com/fluxcd/flux2/issues/3155

⚠️ To be merged after Flux v0.35 release!

### DIFF
--- a/content/en/flux/cheatsheets/oci-artifacts.md
+++ b/content/en/flux/cheatsheets/oci-artifacts.md
@@ -150,7 +150,7 @@ The Flux controllers running on the production cluster
 detects the new semver tag, pulls the manifests and applies them.
 
 {{% alert color="info" title="GitHub Actions" %}}
-If your are using GitHub for CI, please see the [these examples](https://github.com/fluxcd/flux2/tree/main/action) on
+If you are using GitHub for CI, please see [these examples](https://github.com/fluxcd/flux2/tree/main/action) on
 how to automate the publishing of OCI Artifacts in your workflows with the Flux GitHub Action.
 {{% /alert %}}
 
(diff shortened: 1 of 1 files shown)


content/en/flux/cheatsheets/oci-artifacts.md
```diff
@@ -337,6 +337,93 @@ For more details on how to setup contextual authorization for Azure, AWS and Goo
...
+cosign generate-key-pair
+
+kubectl -n flux-system create secret generic cosign-pub \
+  --from-file=cosign.pub=cosign.pub
+```
+
+Push and sign the artifact using the Cosign private key:
+
+```shell
+flux push artifact oci://ghcr.io/stefanprodan/manifests/podinfo:$(git tag --points-at HEAD) \
```
hiddeco, line 362: Maybe replace `ghcr.io/...` here with `<image reference>`? I see it is also used in the examples above, but would think it is easier for people to follow when they have guidance on what needs to be replaced. Instead of accidentally copying the command in full, then discovering they can not push to this image repository, to then figure out what needs replacing.

content/en/flux/cheatsheets/oci-artifacts.md
```diff
@@ -150,7 +150,7 @@ The Flux controllers running on the production cluster
 detects the new semver tag, pulls the manifests and applies them.
 
 {{% alert color="info" title="GitHub Actions" %}}
-If your are using GitHub for CI, please see the [these examples](https://github.com/fluxcd/flux2/tree/main/action) on
+If you are using GitHub for CI, please see the [these examples](https://github.com/fluxcd/flux2/tree/main/action) on
```
darkowlzz, line 153:
```suggestion
If you are using GitHub for CI, please see [these examples](https://github.com/fluxcd/flux2/tree/main/action) on
```

darkowlzz (approved):
LGTM!

PR #1181, opened 2022-10-07
add raffis/gitops-zombies to the tools list

raffis:
Adds a new tool https://github.com/raffis/gitops-zombies to the tools list.

I quickly created a simple tool to identify workload which is not managed by GitOps. 
Quite useful on clusters which are getting migrated to gitops or where pruning is disabled, maybe this helps somebody.


dholbach (commented):
Thanks for the addition! Do you think you could sort the entries alphabetically?

raffis:
> Thanks for the addition! Do you think you could sort the entries alphabetically?

Im AFK this week, will update it asap.

raffis:
> Thanks for the addition! Do you think you could sort the entries alphabetically?

@dholbach done

dholbach (approved):
Nice one, thanks @raffis

PR #1195, opened 2022-10-13
fix helmrelease guide example .sops.yaml file

kingdonb:
This turned out to be the number of backslash escapes needed to get this example to work as a primary source

The error I received without this change from SOPS CLI was: malformed yaml on line 4


pjbgf (approved):
LGTM

PR #1196, opened 2022-10-13
Update flux-e2e.md with info on what extensions are supported by kustomize controller

suryapandian:


kingdonb:
Thanks for adding this. I'm looking for a bit of code or other backup that can verify this – are you sure kustomize controller will only pull these in? (Can we say something about where that decision is made?)

This is good detail to include, but I was always wary about adding fine details like this because they can be changed, and this risks making the doc "less evergreen."

It has been a good amount of time since this doc was broadly refreshed now, it may be time to check if other details are missing or out of date now too.

Anyway, about this detail in particular... if you named a json file in `kustomization.yaml`, I am pretty sure it will get included too, but the `kustomize create --autodetect` command that is mentioned in [this FAQ](https://fluxcd.io/flux/faq/#what-is-the-behavior-of-kustomize-used-by-flux) will not pick it up if you didn't write your own `kustomization.yaml`. I would love to find a piece of source code in either project that says "only `.yml` and `*.yaml` extensions" so we can include a link there in the reference, so it's easy to see if this ever changes out from underneath the docs.

I don't want to make the sentence you added much longer than it is, but it should get some minor adjustment to account for this, and like I said if you know what code filters the list down to only these two extensions, it would be great to include a link to that code directly in the passage! 👍

PR #1201, opened 2022-10-17
Add clomonitor manifest

dholbach:


stefanprodan:
The clomonitor manifest must go under community, that's the repo that CNCF scans.

PR #1210, opened 2022-10-19
FAQ: Should I be using Kustomize remote bases?

stefanprodan:
Encourage users to adopt Flux OCI distribution model and move away from Kustomize remote bases.


dholbach (approved):
Nice one! Thanks @stefanprodan

PR #1211, opened 2022-10-19
FAQ: Should I be using Kustomize Helm chart plugin?

stefanprodan:
Explain why we don't allow the Kustomize Helm plugin and show how Helm works with Flux.

### DIFF
--- a/content/en/flux/faq.md
+++ b/content/en/flux/faq.md
@@ -261,6 +261,66 @@ spec:
       name: staging-cluster
 ```
 
+### Should I be using Kustomize Helm chart plugin?
+
+Due to security and performance reasons, Flux does not allow the execution of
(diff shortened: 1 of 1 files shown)


content/en/flux/faq.md
```diff
@@ -261,6 +261,63 @@ spec:
       name: staging-cluster
 ```
 
+### Should I be using Kustomize Helm chart plugin?
+
+Due to security and performance reasons, Flux does not allow the execution of
+Kustomize plugins which shell-out to random binaries insides its controllers.
```
dholbach, line 267:
```suggestion
Kustomize plugins which shell-out to arbitrary binaries insides its controllers.
```

dholbach (approved):
Thanks for documenting this @stefanprodan

Issue #1219, opened 2022-10-24
[TBD] Remove Flux Legacy docs


dholbach:
:warning: This is still subject of discussion.

Nov 1 we will merge #1218 and even clearer indicate that users will need to update to Flux v2.

If we fully removed docs under `/legacy` it would simplify things in the `website` repository as well. Current modifications because of this:

- 2 separate Algolia search buckets
- overriding of partials
- import of files, etc

Let's discuss.

PR #1218, opened 2022-10-24
Update wording and docs for v1 archival

dholbach:
Signed-off-by: Daniel Holbach <daniel@weave.works>

PR #1220, opened 2022-10-24
Remove Legacy docs (WIP)

dholbach:
Fixes: #1219


dholbach:
Do we want to remove the v1 adopters as well?

dholbach:
I'd be tempted to remove the long "feature parity" section in our roadmap as well.

stefanprodan:
> I'd be tempted to remove the long "feature parity" section in our roadmap as well.

Yes for sure.

dholbach:
With copying the docs back over into the original repositories, this should be a lot less controversial.

dholbach:
This PR is based on top of the changes in #1218.

static/_redirects
```diff
@@ -30,3 +28,5 @@
 
 /docs/contributing/*                            /contributing/:splat                                301!
 /docs/*                                         /flux/:splat                                        301!
+
+/legacy/*                                       /flux/migration                                     301!
```
dholbach, line 32: We could theoretically redirect to GitHub, but that would require a bit of special casing... any strong opinions here?

stefanprodan: I would go with the redirect to migration.

stefanprodan (approved):

static/_redirects
dholbach: Thanks for the feedback!

PR #1221, opened 2022-10-24
Add Pulumi operator as an integration

squaremo:
See also https://www.pulumi.com/blog/pulumi-kubernetes-new-2022/.


stefanprodan (approved):
LGTM

Thanks @squaremo 🏅

PR #1254, opened 2022-11-07
Add Amesto Fortytwo as adoptors

roberthstrand:
Added to both Flux v2 and Flagger list.


stefanprodan (approved):
LGTM

Thanks @roberthstrand

PR #1258, opened 2022-11-13
added automatic scroll and some fixes

Tarunbot:
Signed-off-by: Tarunbot <rajpurohittarun5@gmail.com>


hiddeco (commented):
Think this visually looks good, would maybe just make the scrolling a tiny bit slower?

PR #1287, opened 2022-11-29
update build to python 3.11 (current)

dholbach:


dholbach:
Oops, 3.8 is current in Netlify: https://github.com/netlify/build-image/blob/focal/included_software.md#tools

PR #1290, opened 2022-11-30
Add template for monthly update blog posts

dholbach:
This will be an experiment to see if we can spare ourselves the pain of going from Google Docs to .docx to .md for this very regular set of communications.


dholbach:
In a next step, we can have the script update some of the generic bits in the post as well, like YEAR, MONTH and the like.

PR #1308, opened 2022-12-07
Persistent storage for Flux internal artifacts

stefanprodan:
Add instructions on how to configure persistent storage for source-controller data.


dholbach (approved):
Thanks @stefanprodan

Issue #870, opened 2022-04-05
Make support page more inspiring


dholbach:
Right now the `/support` comes from <https://github.com/fluxcd/community/blob/main/SUPPORT.md>

It'd be great if the companies who offer support could showcase their work and services somehow like <https://linkerd.io/support-training/>

tarunrsl:
Hello I want to work on this feature _ is it still Open to work?

Tarunbot:
Hey @dholbach  i am interested to make the support page as powerful as the mentioned one so assign me this issue so that i can resolve it .Thank you:)

dholbach:
That's great - thanks @Tarunbot ... let me know if I can help in any way. If you have ideas on how it could be done, maybe let's talk in #flux-contributors on CNCF Slack about it?

UnfixedSun:
Hey @Tarunbot @dholbach please assign me this issue. Thanks.

dholbach:
@UnfixedSun and @Tarunbot Maybe you can collaborate on this issue?

Tarunbot:
@dholbach  ohkie we will thank you:)

Tarunbot:
@dholbach  hey currently i am working on some other issue so wouldnt be able to work on it so u can assign this to someone else Thank you

h20220025:
hey, @dholbach I would like to work on this issue, I am looking forward to it. Could you assign this issue to me?

dholbach:
I'm starting work on this as well and will add a draft PR in a bit. If you want to collaborate, let's do it!

PR #1318, opened 2022-12-12
Beautify support page

dholbach:
Ref: #870


dholbach:
Correct - my aim was 

- make page more inspiring
- drop legacy mentions
- make headings clearer

It's a good step forward in closing #870

PR #1351, opened 2023-01-05
FLUX DOCS: don't use legacy slack config in example

digitalbanana:
Signed-off-by: Dave Lowe <dave@digitalbanana.co.uk>

proposed change to use the recommended slack integration (not legacy)

### DIFF
--- a/content/en/flux/guides/notifications.md
+++ b/content/en/flux/guides/notifications.md
@@ -27,15 +27,17 @@ The notification controller is part of the default Flux installation.
 
 ## Define a provider
 
-First create a secret with your Slack incoming webhook:
+First create a secret with your Slack Bot token:
 
 ```sh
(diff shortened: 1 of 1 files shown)


kingdonb:
Thanks, can you squash your commits @digitalbanana 

I think I noticed this issue when I was setting up notifications last time (the legacy webhook is no longer an option to create, so you have to use the Bot token, but nothing else changes in these instructions),

Thank you for making a PR to close it 🎉

kingdonb:
It looks like you have noticed some things I didn't spot myself, actually

Please squash and let me know when you are finished, then we can approve it 👍

kingdonb (changes requested):
Squash please

digitalbanana:
obsolete - didnt get round to this :-(

PR #1353, opened 2023-01-11
Add FAQ section regarding overriding HelmRelease values with kustomization

lassizci:
I was looking for a way to set helm chart (or image tag)  version separately for each cluster, following the structure in: https://github.com/fluxcd/flux2-kustomize-helm-example.

https://pkg.go.dev/github.com/fluxcd/pkg/apis/kustomize#Image seemed simple for overriding the image, although not the helm chart version, but it doesn't actually work, since we are making kustomization to `HelmRelease`. So I though something like this in the FAQ could be helpful. What do you think?

### DIFF
--- a/content/en/flux/faq.md
+++ b/content/en/flux/faq.md
@@ -454,6 +454,103 @@ EOF
 Based on the above definition, Flux will upgrade the release automatically
 when Bitnami publishes a new version of the metrics-server chart.
 
+### How do I set local overrides to a Helm chart?
+
+Lets assume we have a common `HelmRelease` definition we use as a base and we
+we need to further customize it e.g per cluster, tenant, environment and so on:
+```yaml
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: podinfo
+  namespace: podinfo
+spec:
+  releaseName: podinfo
+  chart:
+    spec:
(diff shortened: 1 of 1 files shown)


content/en/flux/faq.md
```diff
@@ -454,6 +454,103 @@ EOF
...
+        kind: HelmRepository
+        name: podinfo
+  interval: 50m
+  install:
+    remediation:
+      retries: 3
+```
+
+And then we would specify `postBuild` in our `kustomization`:
+```yaml
```
hiddeco, line 533:
```suggestion
To enable the replacement of the `PODINFO_CHART_VERSION` variable with a different version (than the `6.2.0` default), specify `postBuild` in our `Kustomization`:

```yaml
```

Is a bit more explanatory in terms of what would happen, and sheds some light on the `:=6.2.0`.

lassizci:
> Thank you @lassizci, think this could indeed be useful. Made a couple of suggestions on your draft.

Thank you for the feedback! I tried to make the "per cluster" case a bit more generic and addressed the rest of the points. Could you have another look?

makkes (approved):
Great entry to our FAQ. Thanks! What I've been using sometimes is use a ConfigMap for holding the substitution values but I suppose this might overload the FAQ entry a little so I'm fine with how it is.

PR #1362, opened 2023-01-30
landing page: remove terminal for now

dholbach:
Remove terminal for now.


dholbach:
→ https://deploy-preview-1362--fluxcd.netlify.app/

stefanprodan:
We need to update the terminal as it shows Flux bootstrap from one or two years ago, things have changed a lot I think.

dholbach:
@hiddeco Do you think you can help with this? I'm not sure how this was done or set up...

stefanprodan:
I also see a major improvement on iOS with the current home page, the slowness seems to be due to the terminal rendering.

dholbach:
If we feel we are good to go without the terminal, there's quite a lot of code we can remove from `f/website`.

hiddeco:
As I do not have time at present to re-do the whole recording, in combination with Stefan's comments about mobile problems etc. I am fine with removing it for now.

kingdonb (approved):
👍

PR #1374, opened 2023-02-13
image-reflector-controller v1beta2 updates

darkowlzz:
- Update the mention of old API version.
- Add example of the latest ten tags in the scan result.
- Remove section about authentication with old autologin flags and add
  reference to the spec docs provider section.
- Add ImagePolicy examples section that are no longer in the spec docs.
- Update link to native auth docs and remove empty section about GCP
native auto-login.


stefanprodan (approved):
LGTM

Thanks @darkowlzz 🏅

hiddeco (approved):

PR #1375, opened 2023-02-13
Update explanation on Helm use case 

somtochiama:
The explanation after the `flux create source helm` command seems to imply that the command will commit the files to git.
This pull request tries to fix that.

Ref: https://cloud-native.slack.com/archives/CLAJ40HV3/p1676320569584019
Signed-off-by: Somtochi Onyekwere <somtochionyekwere@gmail.com>

### DIFF
--- a/content/en/flux/use-cases/helm.md
+++ b/content/en/flux/use-cases/helm.md
@@ -48,21 +48,21 @@ helm install my-traefik traefik/traefik \
 Flux client:
 
 ```sh
-flux create source helm traefik --url https://helm.traefik.io/traefik --namespace traefik
+flux create source helm traefik --url https://helm.traefik.io/traefik --namespace traefik --export > /flux/boot/traefik/helmrepo.yaml
 flux create helmrelease my-traefik --chart traefik \
   --source HelmRepository/traefik \
   --chart-version 9.18.2 \
-  --namespace traefik
+  --namespace traefik \
+  --export > /flux/boot/traefik/helmrepo.yaml
 ```
 
-The main difference is that the Flux client will not imperatively create resources in the cluster.
-Instead, these commands create Custom Resource *files*, which are committed to version control
-as instructions only (note: you may use the `--export` flag to manage any file edits with
-finer grained control before pushing to version control).
-Separately, the Flux Helm Controller automatically reconciles these instructions
-with the running state of your cluster based on your configured rules.
+These commands save the YAML for the Flux helm custom resources to a file which can be edited and pushed to git.
+When these resources are pushed to a repository that Flux watches, Flux applies them on the cluster and the Flux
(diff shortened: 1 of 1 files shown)


somtochiama:
@stefanprodan I don't know if we want to add an beginning section where we say we expected them to have flux bootstrapped

stefanprodan:
@somtochiama yes, we need to tell people to bootstrap and link to the install doc.

content/en/flux/use-cases/helm.md
```diff
@@ -47,25 +54,26 @@ helm install my-traefik traefik/traefik \
...
+
 ```sh
-flux create source helm traefik --url https://helm.traefik.io/traefik --namespace traefik --export > /flux/boot/traefik/helmrepo.yaml
+flux create source helm traefik --url https://helm.traefik.io/traefik --namespace traefik --export > traefik/helmrepo.yaml
 flux create helmrelease my-traefik --chart traefik \
   --source HelmRepository/traefik \
   --chart-version 9.18.2 \
   --namespace traefik \
-  --export > /flux/boot/traefik/helmrepo.yaml
+  --export > traefik/helmrepo.yaml
```
stefanprodan, line 65: Can you please test this with Kind, I guess the HR will fail due to the lack of LB support.

stefanprodan: I propose we replace traefik with nginx, and then we can use the HR from here which is e2e tested on Kind. https://github.com/fluxcd/flux2-kustomize-helm-example/blob/main/infrastructure/controllers/ingress-nginx.yaml

somtochiama: @stefanprodan you are right
```
 Helm install failed: timed out waiting for the condition                                                                                                                                │
                                                                                                                                                                                              │
      Last Helm logs:                                                                                                                                                                         │
                                                                                                                                                                                              │
      creating 1 resource(s)                                                                                                                                                                  │
      CRD traefikservices.traefik.containo.us is already present. Skipping.                                                                                                                   │
      creating 5 resource(s)                                                                                                                                                                  │
      beginning wait for 5 resources with timeout of 5m0s                                                                                                                                     │
      Service does not have load balancer ingress IP address: flux-system/my-traefik
```

I will update it to use nginx

somtochiama: Nginx also has the same problem. Should we use podinfo?

stefanprodan: Nginx does not have this issue, see the values used in the link I posted above

somtochiama: In the usecase guide, values are introduced in a later section after the installation. 
So I was thinking we could start with something that just works without tweaking the values first

stefanprodan: OK lets' use Flagger.

PR #1386, opened 2023-02-20
Disable Kubernetes cluster role aggregations at bootstrap

stefanprodan:
Ref: https://github.com/fluxcd/flux2/discussions/3616


dholbach (approved):
Thanks @stefanprodan

Issue #1402, opened 2023-03-06
Landing page: too broad for small mobile screens


dholbach:
![grafik](https://user-images.githubusercontent.com/1346979/223105218-f20cf933-76b2-457b-9e1e-1086a2f327fc.png)

Currently the landing page is too broad for small mobile screens, it looks like it's the calendar extending too far.

PR #1404, opened 2023-03-06
fix calendar block on mobile

Arhell:
resolve https://github.com/fluxcd/website/issues/1402


dholbach (approved):
Thanks for fixing this!

Issue #1405, opened 2023-03-07
adopters Weaveworks goes beyond the borders


Arhell:
![Screenshot_1](https://user-images.githubusercontent.com/26163841/223569811-d3b2d08f-f821-4815-95b8-39a4ebd08f65.png)
i will fix

PR #1407, opened 2023-03-08
fix adopters Weaveworks block

Arhell:
resolve https://github.com/fluxcd/website/issues/1405


stefanprodan (approved):
LGTM

Thanks @Arhell 🥇

PR #1419, opened 2023-03-14
Update to new docsy and new bootstrap (WIP)

dholbach:
This is still work-in-progress and the new docsy has not been released just yet. As it requires us to update our own partials and styling to the new world (bootstrap 4 will EOL this year) and re-adapt the changes we have on top of docsy, it will take a while and we are already starting this work in this PR.


dholbach:
- [ ] Wait for new docsy to be released
- [ ] Three-way merge navbar partial
- [ ] Check we have all partials updated
- [ ] Review CSS changes
- [ ] Check https://getbootstrap.com/docs/5.2/migration/
- [ ] Review https://github.com/google/docsy/blob/main/CHANGELOG.md#070---next-planned-release-unpublished-yet

dholbach:
Somebody will need to take over this PR, as I won't have time to finish it.

PR #1421, opened 2023-03-20
Add documentation for new `--deploy-token-auth` option

timofurrer:
This change documents the possibility to use a deploy token instead of a personal access token or an SSH deploy key.

This feature is implemented with:

* https://github.com/fluxcd/flux2/pull/3654

and was discussed in:

* https://github.com/fluxcd/flux2/discussions/3595

/cc @makkes

### DIFF
--- a/content/en/flux/installation.md
+++ b/content/en/flux/installation.md
@@ -284,6 +284,18 @@ flux bootstrap gitlab \
   --personal
 ```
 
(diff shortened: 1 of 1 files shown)


makkes (approved):
🥇 let's merge it once fluxcd/flux2#3654 is merged

makkes (approved):

PR #1427, opened 2023-03-23
GA: Publish v1 docs for GitRepository, Receiver & Kustomization

makkes:
Changes:
- adapt docs script to...
  - take into account the v1 split
  - consume `weight` comments for proper ordering of pages
- add both v1 and v1beta2 generated docs under "References" section
- update references to old API versions on all pages


stefanprodan:
It looks great! Thanks @makkes 

I've put this on hold until we release NC, then you can switch back to semver in the script.

stefanprodan (approved):
LGTM

Thanks @makkes 🥇

PR #1438, opened 2023-03-30
Fix Azure SOPS guide commands

paddatrapper:


content/en/flux/guides/mozilla-sops.md
```diff
@@ -362,11 +362,11 @@ Create an `AzureIdentity` object that references the identity created above:
 apiVersion: aadpodidentity.k8s.io/v1
 kind: AzureIdentity
 metadata:
-  name: sops-akv-decryptor  # kustomize-controller label will match this name
+  name: ${IDENTITY_NAME}  # kustomize-controller label will match this name
```
somtochiama, line 365: Hey @paddatrapper ,
Thanks for this. The name here doesn't have to be the same as the `${IDENTITY_NAME}`  name so we can leave it as-is

somtochiama (changes requested):

content/en/flux/guides/mozilla-sops.md
paddatrapper: The instructions later, e.g. `AzureIdentitBinding` use `${IDENTITY_NAME}` in the selector and the patch of the `Deployment`, so I renamed this to be consistent with the later manifests

somtochiama (approved):
LGTM!

Thanks, @paddatrapper 🎉

stefanprodan (approved):

PR #1441, opened 2023-04-01
Update gh-actions-manifest-generation.md

kwongtn:
Currently it links to `https://fluxcd.io/guides/sortable-image-tags/`, which is incorrect.
It should link to `https://fluxcd.io/flux/guides/sortable-image-tags/` instead


kingdonb:
Where are you seeing the broken link exactly?

image

I'm not sure what happened with the build failure, I don't think it was caused by your change. But I also don't see the bad link.

kwongtn:
@kingdonb that's real weird, previously when I clicked it gives me the incorrect link, but now when I checked back it is correct -- probably error on another page 🤔
Closing this issue for the time being, sorry for the inconvenience 😅

PR #1458, opened 2023-04-12
Use flux_bootstrap_git in Terraform docs

stefanprodan:
Part of: https://github.com/fluxcd/flux2/issues/3735

### DIFF
--- a/content/en/flux/installation.md
+++ b/content/en/flux/installation.md
@@ -462,26 +462,35 @@ with `flux --version`.
 
 The bootstrap procedure can be implemented with Terraform using the Flux provider published on
 [registry.terraform.io](https://registry.terraform.io/providers/fluxcd/flux).
+The provider offers a Terraform resource called
+[flux_bootstrap_git](https://registry.terraform.io/providers/fluxcd/flux/latest/docs/resources/bootstrap_git)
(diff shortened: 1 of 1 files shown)


content/en/flux/installation.md
```diff
@@ -462,26 +462,35 @@ with `flux --version`.
...
+  }
 }
 
-data "flux_sync" "main" {
-  target_path = "clusters/my-cluster"
-  url         = "https://github.com/${var.github_owner}/${var.repository_name}"
-  branch      = "main"
+resource "flux_bootstrap_git" "this" {
+  path                   = "clusters/my-cluster"
+  network_policy         = true
```
phillebaba, line 487: Network policy is true by default.

content/en/flux/installation.md
```diff
@@ -462,26 +462,35 @@ with `flux --version`.
...
+  kustomization_override = file("${path.module}/kustomization.yaml")
```
phillebaba, line 488: What is the contents of kustomization_override?

content/en/flux/installation.md
stefanprodan: I know, this is here to showcase that it can be disabled.

PR #1461, opened 2023-04-12
Update kuma tutorial for flagger docs

aryan9600:
sync changes made in https://github.com/fluxcd/flagger/pull/1402


stefanprodan:
Is this the only change to docs since the last sync?

stefanprodan (approved):

Issue #1420, opened 2023-03-14
Core concepts: consider explaining GitOps visually


dholbach:
We got feedback that it might be a good idea to explain GitOps visually, like in a diagram.

@errordeveloper pointed out https://www.weave.works/blog/kubernetes-anti-patterns-let-s-do-gitops-not-ciops as an example.

FavourDaniel:
Working on this...

PR #1462, opened 2023-04-14
Added GitOps Workflow Image

FavourDaniel:
Added GitOps Workflow image showing how it works with Flux to resolve issue #1420.

### DIFF
--- a/content/en/flux/concepts.md
+++ b/content/en/flux/concepts.md
@@ -13,6 +13,8 @@ is described declaratively and version controlled (most likely in a Git reposito
 and having an automated process that ensures that the deployed environment
 matches the state specified in a repository.
 
+![GitOps Workflow](https://user-images.githubusercontent.com/89241109/232053334-5c5f29e7-0063-488a-86af-9bc692b545f2.png)
+
 For more information, take a look at ["What is GitOps?"](https://www.gitops.tech/#what-is-gitops).
 
 ## Sources


stefanprodan:
Hey @FavourDaniel the following two weeks I'll be off, first week is KubeCon then I'm going on vacation. Let's discuss this when I get back please. I suggest we use https://excalidraw.com and we upload the source file in the repo, so anyone can modify it. Also the diagram is confusing to me, Flux doesn't pull images, it only lists tags, also that step with the manifest doesn't look good IMO, we need to incorporate it under the repo.

FavourDaniel:
> Hey @FavourDaniel the following two weeks I'll be off, first week is KubeCon then I'm going on vacation. Let's discuss this when I get back please. I suggest we use https://excalidraw.com and we upload the source file in the repo, so anyone can modify it. Also the diagram is confusing to me, Flux doesn't pull images, it only lists tags, also that step with the manifest doesn't look good IMO, we need to incorporate it under the repo.

Alright, I will close this PR and make changes, thank you. Enjoy your vacation!

stefanprodan:
I'm working on diagrams today for my KubeCon talk, here is one made with excalidraw: 

Screenshot 2023-04-14 at 15 54 51

Do you like this design?

FavourDaniel:
> I'm working on diagrams today for my KubeCon talk, here is one made with excalidraw:
> 
> Screenshot 2023-04-14 at 15 54 51
> 
> Do you like this design?

Yes, this really good, love it. I'll try to replicate this design style with Excali as well, I'll reach out to you on Slack now.

PR #1464, opened 2023-04-17
Add Orange to adopters

davidblaisonneau-orange:
As Orange is using FluxCD to deploy network functions, we propose you to be part of the adopters.

Thanks to all developer behind FluxCD !


hiddeco (approved):

hiddeco:
Thank you very much @davidblaisonneau-orange :bow:

stefanprodan (approved):
💯

Issue #718, opened 2022-01-25
Reconsider migration guides placement


dholbach:
From the CNCF Tech Docs assessment of our docs (cncf/techdocs#89):

>    - The [Migration](https://fluxcd.io/docs/migration/) section isn't more important than [Toolkit components](https://fluxcd.io/docs/components/), [Toolkit dev guides](https://fluxcd.io/docs/gitops-toolkit/), or the [Flux CLI reference](https://fluxcd.io/docs/cmd/). Placing it above these in the information architecture doesn't make sense. 
> 
> ### Recommendations
> 
> - Move the [Migration Guide](https://fluxcd.io/docs/guides/) to either a subsection of [Get Started](https://fluxcd.io/docs/get-started/) or a subsection of [Installation](https://fluxcd.io/docs/installation/)
>    - Consider renaming "Migration guide" to "Migrate from Flux v1" for clarity

Neilblaze:
@dholbach is this open to contributions? If yes, I'd like to raise a PR regarding it.

dholbach:
This issue is linked with #717

Hammadh7:
I am interested in this issue

stefanprodan:
Having migration under installation would confuse most Flux users, as this about migrating from a deprecated version (v1) and not about migrating between v2 releases. We'll probably have a migration section, for v2 APIs GA, alpha to stable controllers and so on.

PR #1466, opened 2023-04-17
new migration.md

Hammadh7:
#718
I have added the migration section inside the Installation section


dholbach (changes requested):
It looks like you are including a new file, which has no markup and copied parts of a GitHub page?

PR #1488, opened 2023-05-03
File name mismatch between the error and example's spec

albertomr86:
Match the file name that appears in the error with the example.

image


makkes:
Thanks @albertomr86!

PR #1491, opened 2023-05-04
Add blablacar to adopters

jhaumont:
Hello

We have forgot to update the website but we used Flux v2 in production since July 2022 & Flagger since December 2021 (with Flux v1 at first)
Thanks for the job and we are always available to discuss if you want to have feedbacks from us or evaluate new features.

Let's go to OCIOps 😄 

EP team @blablacar


makkes (approved):
🎉

PR #1499, opened 2023-05-12
Add Workload Identity to Azure guide

tiffanywang3:


content/en/flux/use-cases/azure.md
```diff
@@ -54,6 +71,7 @@ az aks create \
  --network-policy="calico" \
  --enable-managed-identity \
  --enable-pod-identity \
+ --enable-workload-identity \
```
stefanprodan, line 74: Do both flag work? I expect the command to fail. Better set only `--enable-workload-identity`

stefanprodan (approved):
LGTM

Thanks @tiffanywang3 🏅

PR #1501, opened 2023-05-15
Correct gcp image automation link in gcp source repo use-case

heubeck:


stefanprodan (approved):
LGTM

Thanks @heubeck for the fix

PR #1502, opened 2023-05-22
docs: fix GitHub Pull Request Action configuration syntax

cybercyst:
`create` isn't valid syntax and tries to create a PR for every branch pushed


content/en/flux/use-cases/gh-actions-auto-pr.md
```diff
@@ -32,7 +32,7 @@ To create the pull request whenever automation creates a new branch, in your man
 # ./.github/workflows/staging-auto-pr.yaml
 name: Staging Auto-PR
 on:
-  create:
+  push:
```
makkes, line 35: This would run this workflow whenever commits are pushed to this branch which is also not desirable. I think the only way to accomplish this is to use the `create` event and add appropriate `if` conditions to the jobs to check the ref name.

makkes (changes requested):

PR #1507, opened 2023-05-29
Update GitHub Actions examples syntax + actions versions

Antvirf:
Primary purpose is to update the GH Actions syntax in preparation for the [deprecation of `set-output`](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), in order to keep the examples functional in the future.

In addition to that, I made some straightforward updates for the well-known GH Actions used in the examples like `docker/login-action`, `docker/setup-*` and `actions/checkout`.


stefanprodan (approved):
LGTM 

Thanks @Antvirf 🏅

PR #1522, opened 2023-06-13
docs: updated zero-downtime-deployments.md

eabykov:
1. Removed `wget` HTTP probes:
    - Istio solves both these problems by rewriting the application PodSpec readiness/liveness probe, so that the probe request is sent to the [sidecar agent](https://istio.io/latest/docs/reference/commands/pilot-agent/). For HTTP and gRPC requests, the sidecar agent redirects the request to the application and strips the response body, only returning the response code. For TCP probes, the sidecar agent will then do the port check while avoiding the traffic redirection. https://istio.io/latest/docs/ops/configuration/mesh/app-health-check/
    - Linkerd automatically authorizes probes to pods https://linkerd.io/2.13/features/server-policy/#default-authorizations
2. Added instruction how to hold application until Service Mesh Sidecar starts
3. Outdated `Delay Envoy shutdown` replaced with `Delay Service mesh proxy shutdown`
4. Updated HPA api version to `autoscaling/v2` that available since v1.23

### DIFF
--- a/content/en/flagger/tutorials/zero-downtime-deployments.md
+++ b/content/en/flagger/tutorials/zero-downtime-deployments.md
@@ -28,19 +28,12 @@ You application should expose a HTTP endpoint that Kubernetes can call to determ
 
 ```yaml
 livenessProbe:
-  exec:
-    command:
-    - wget
-    - --quiet
-    - --tries=1
-    - --timeout=4
-    - --spider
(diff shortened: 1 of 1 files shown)


agbykov (approved):

eabykov:
@stefanprodan please take a look

eabykov:
@darkowlzz please take a look

PR #1523, opened 2023-06-13
Update aws-codecommit.md

gopikrishna7:
Added info regarding creating source from the AWS CodeCommit


gopikrishna7:
@stefanprodan Please review

PR #1526, opened 2023-06-22
openshift: Explain how to remove the pod-security labels

sastorsl:
OpenShift since 4.11 adds the same pod-security labels to the `flux-system` Namespace as flux defines, but with other values.
This triggers endless reconciliations.

This reconciliation can be avoided by *removing* the label specifications from flux - with a patch to the `Kustomization` object.
This PR documents how this can be done.

Example

```
kustomization/flux-system.flux-system
Namespace/flux-system configured
```

Fix: https://github.com/fluxcd/flux2/issues/3995


stefanprodan (approved):
LGTM

Thanks @sastorsl

PR #1528, opened 2023-06-23
Security: Recommend sharding for resource isolation

stefanprodan:
For resource isolation and improved reliability we now recommend users to configured Flux sharding instead of fiddling with the namespace watcher which was never meant to be used outside internal testing.

### DIFF
--- a/content/en/flux/security/best-practices.md
+++ b/content/en/flux/security/best-practices.md
@@ -17,8 +17,9 @@ convenience, performance or resources utilization of Flux. Therefore, use this i
 combination with your own Security Posture and Risk Appetite.
 
 Some recommendations may overlap with Kubernetes security recommendations, to keep
-this short and more easily maintainable, please refer to [Kubernetes CIS Benchmark](https://www.cisecurity.org/benchmark/kubernetes) for non
-Flux-specific guidance.
+this short and more easily maintainable, please refer to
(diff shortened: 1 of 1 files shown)


content/en/flux/security/best-practices.md
```diff
@@ -207,23 +208,23 @@ The recommendations below are based on Flux's latest version.
...
+    Running multiple Flux instances within the same cluster is supported by means of sharding, please consult the [Flux sharding and horizontal scaling documentation](flux/cheatsheets/sharding/) for more details.
 
-    In order for controllers not to compete with each other when trying to reconcile Custom Resources, controller types (e.g. `image-automation-controller`) must have their `--watch-all-namespaces` setting aligned across the cluster. If one instance has that flag set to `false`, all other instances must be set to the same value. When it is set to true (e.g. `--watch-all-namespaces=true`), there must be a single instance of that controller type within the cluster.
+    In order for controllers not to compete with each other when trying to reconcile Custom Resources, controller types (e.g. `source-controller`) must have be configured with unique label selectors in the `--watch-label-selector` flag.
```
hiddeco, line 222:
```suggestion
    To avoid conflicts among controllers while attempting to reconcile Custom Resources, controller types (e.g. `source-controller`) must have be configured with unique label selectors in the `--watch-label-selector` flag.
```

maybe?

hiddeco (approved):

PR #1551, opened 2023-07-07
Fix 404 page

makkes:
The page isn't rendered properly (completely empty) with that comment at the top.

Before: https://fluxcd.io/does-not-exist

After: https://deploy-preview-1551--fluxcd.netlify.app/does-not-exist

### DIFF
--- a/assets/scss/_page.scss
+++ b/assets/scss/_page.scss
@@ -24,6 +24,11 @@
 }
 
 .page404 {
+  .DocSearch-Button {
+    width: 100%;
+    margin-left: 0;
+  }
+
   h1 {
     font-size: 2rem;
     font-weight: 700;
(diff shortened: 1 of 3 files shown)


hiddeco:
In the preview the search doesn’t appear to actually work for me.

makkes:
> In the preview the search doesn’t appear to actually work for me.

Oh right. I already know why. On it.

assets/scss/_page.scss
```diff
@@ -24,6 +24,12 @@
 }
 
 .page404 {
+
```
hiddeco, line 27:
```suggestion
```

hiddeco (approved):

hiddeco:
Nice work @makkes! :bow:

PR #1555, opened 2023-07-07
Added Sonatype to adopters

eddie-knight:


stefanprodan (approved):
Thank you @eddie-knight 🏅

Issue #1556, opened 2023-07-09
update hugo to v0.115.2


Arhell:
https://github.com/gohugoio/hugo/releases/tag/v0.115.2

Issue #1390, opened 2023-02-26
Docker Build action is missing some modernities


kingdonb:
I'm dogfooding some of my docs, and this one has gone a bit stale:

https://fluxcd.io/flux/use-cases/gh-actions-app-builder/

It needs at least:

```
! The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
docker: .github#248
```

and

```
permissions:
  packages: write
```

from a first pass – I'll come back to this later, but right now I just wanted to be sure I don't lose track from this because I'm moving on just this minute...

PR #1557, opened 2023-07-09
update hugo to v0.115.2

Arhell:
resolve https://github.com/fluxcd/website/issues/1556

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,7 +3,7 @@
   command = "make production-build"
 
 [build.environment]
-  HUGO_VERSION = "0.115.1"
+  HUGO_VERSION = "0.115.2"
   NODE_VERSION = "16"
   GO_VERSION = "1.18"
 


makkes (approved):
I wonder if we should just automate the PR creation. 🤔

Arhell:
> I wonder if we should just automate the PR creation. 🤔

maybe, but it's not an dependencies. 

dependencies [here](https://github.com/fluxcd/website/blob/main/.github/dependabot.yml), and so I make my small contribution

kingdonb:
We should really automate something though, and that is the rebuild of the Docker image, which is meant to be done every time hugo upgrade happens else some of the FluxCD website developer guide stops working (ostensibly, I'm not actively running it to test...)

I thought I had an issue tracking this, but I only found #1390 

Since I guess I haven't documented this idea anywhere else... I'll update 1390 with the details of what I'm asking about!

PR #1558, opened 2023-07-10
Set sensible interval values

stefanprodan:
Various reconcile intervals were set too low, this PR sets sensible interval values so users don't start on the wrong foot and DDOS the various APIs Flux integrates with.


kingdonb (approved):
👍

makkes (approved):
👏🏻

Issue #1560, opened 2023-07-11
[GSoD] Reorganise the installation docs  


stefanprodan:
This is a proposal to reorganise the [installation](https://fluxcd.io/flux/installation/) documentation with the goal of having all the Flux install options and the various bootstrap configurations (Git providers, Cloud providers, controller customisations, multi-tenancy etc) in one place.

The proposal is to split the current page into sections (per Git provider) and to bring together the installation specific information from Cheatsheets ([bootstrap](https://fluxcd.io/flux/cheatsheets/bootstrap/) and [sharding](https://fluxcd.io/flux/cheatsheets/sharding/)) and Use Cases ([azure](https://fluxcd.io/flux/use-cases/azure/), [aws](https://fluxcd.io/flux/use-cases/aws-codecommit/), [gcp](https://fluxcd.io/flux/use-cases/gcp-source-repository/)) under the installation section.

We'll need to setup redirects for Cheatsheets and Use Cases to limit the impact of these changes. Given that links to these docs are used in hundreds of GH issues, discussions, PRs, Blog posts, Slack threads, StackOverflow, etc we need to make sure people don't end up on a 404 page.

PR #1564, opened 2023-07-11
Mark some use-cases docs stale

kingdonb:
If anyone thinks that one of these does not need any updates, then we can remove it from this list pre-emptively.

I added a notice to each doc that I haven't personally reviewed recently, (regardless of whether I am the original author. Trying to use a broad stroke to get as much done as possible.)

The use-cases docs that are not covered here are all ones intended to be migrated into the Installation guides, per the GSoD plan that is being advanced here. So the use-cases section of the docs may go away entirely, (unless some of these docs turn out to survive all this without getting relocated):

https://github.com/fluxcd/website/issues/1560

### DIFF
--- a/content/en/flux/flux-e2e.md
+++ b/content/en/flux/flux-e2e.md
@@ -5,6 +5,12 @@ description: "A narrative of the life of a commit as it relates to Flux componen
 weight: 146
 ---
 
+{{% alert color="warning" title="Disclaimer" %}}
+Note that this guide has not been updated since more than a year ago, it does not address `OCIRepository`, and needs to be refreshed for Flux 2.0.0.
+
+Expect this doc to receive an overhaul soon.
+{{% /alert %}}
+
 Below we describe the flow of data through Flux, from End to End.
 
 We assume a standard Flux installation, with all optional features enabled, then explain how Flux users can expect their change to flow through the stages
--- a/content/en/flux/flux-e2e.md
+++ b/content/en/flux/flux-e2e.md
@@ -6,7 +6,7 @@ weight: 146
 ---
 
 {{% alert color="warning" title="Disclaimer" %}}
-Note that this guide has not been updated since more than a year ago, it does not address `OCIRepository`, and needs to be refreshed for Flux 2.0.0.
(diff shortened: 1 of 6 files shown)


kingdonb:
I will double-check the formatting, open to suggestions on how to word these notices better. I tried to be as specific as I could, but the docs which I actually just haven't reviewed yet notices are pretty generic and basic text to that effect.

kingdonb:
I missed the OpenShift and Helm guides, they most likely belonged on this list as well

kingdonb:
> Note that this guide needs review in consideration of Flux v2, and likely needs to be refreshed.
> Expect this doc to either be archived soon, or to receive some overhaul.

This is the basic text that unreviewed docs are getting. (We can also simply do the review of docs on the list ahead of merging this PR, and take them off the list then.)

The [GitHub Actions Auto PR doc](https://fluxcd.io/flux/use-cases/gh-actions-auto-pr/) probably doesn't need many or any changes.

kingdonb:
The Flux from End-to-End doc should be included in this as well:

https://fluxcd.io/flux/flux-e2e/

It has yet to receive updates for any OCIRepository changes. It's time for an overhaul of this doc. (At least it doesn't mention implementation details like libgit2...)

content/en/flux/use-cases/gh-actions-helm-promotion.md
```diff
@@ -5,6 +5,12 @@ description: "How to configure a promotion workflow for Flux HelmReleases with G
 weight: 31
 ---
 
+{{% alert color="warning" title="Disclaimer" %}}
```
stefanprodan, line 8: This guide was updated for Flux 2.0.0.

stefanprodan (changes requested):

content/en/flux/use-cases/gh-actions-helm-promotion.md
kingdonb: OK, I fixed this in e2fdb17f905aca6c190e14eed746780883123ad8

stefanprodan (approved):
LGTM 

Thanks @kingdonb

PR #1571, opened 2023-07-19
Fix ordering of versions when pulling component docs

makkes:
`sort -V` doesn't properly sort according to semver. This led to an old version of NC (1.0.0-rc.4) being pulled instead of the latest 1.0.0.


stefanprodan (approved):
LGTM

Thanks @makkes 🏅

PR #1575, opened 2023-07-20
fix regex example in flux notification for helm release promotion

fparga:
Fixed the regex shown in the example for the Alert/Notification used to trigger an helm release promotion GitHub workflow.

The example that was there give the following error:

> error parsing regexp: missing argument to repetition operator: *


stefanprodan (approved):
LGTM

Thanks @fparga

PR #1582, opened 2023-07-24
Add EcoVadis to adopters

AliakseiVenski:
We've been using Flagger for 6 month in production already. Works really great. Thank you 😄


stefanprodan (approved):
Awesome thanks @AliakseiVenski

PR #1583, opened 2023-07-26
Add golang to hugo-support image

kingdonb:
We've had this docker serve stuff in the Makefile and it has been broken since some time ago, I am working it out in this branch.

### DIFF
--- a/.github/workflows/build-push-hugo-image.yml
+++ b/.github/workflows/build-push-hugo-image.yml
@@ -12,6 +12,7 @@ jobs:
         uses: docker/setup-qemu-action@v2
       -
         name: Set up Docker Buildx
+        id: buildx
         uses: docker/setup-buildx-action@v2
       -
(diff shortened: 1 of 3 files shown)


kingdonb:
I was able to build and run the environment locally again with these changes. The docs did not appear to need any updates just now, but the runtime environment apparently requires Go now (or will soon) which I added in the support image.

makkes (commented):
This lgtm but I'm leaning towards putting the changes to the Makefile into #1577 because without any context it's not easily understandable why the `BRANCH` variable is introduced here. wdyt?

kingdonb:
Yes, I'll pull them out from this branch so we can merge it, (the commit id was 1aaf55f so we do not lose track)

kingdonb:
Should be in shape for merge 👍

makkes (approved):

PR #1584, opened 2023-07-27
Arm64 arch in website dev environment, path-based trigger

kingdonb:
This change was made on the build-hugo workflow in #1583, but we will need to make the change here as well


kingdonb:
When this build passes, we'll have multi-arch Hugo images:

https://github.com/fluxcd/website/actions/runs/5681642492

The hugo-support image addressed by this PR, uses `Dockerfile` in the local repository, and represents all of our own requirements atop of Hugo itself. I've tested all of these builds locally, the only thing that remains here is I think to update the remaining GHA workflow, and run the image builds by hand.

A separate PR might still add the capability to automatically trigger the necessary hugo and hugo-support builds on change to any of the files that can necessitate a rebuild. I haven't worked all that out yet, we've only needed these images a very few times (as most of the needs for testing are already being met by the Netlify preview feature!)

kingdonb:
I tested the consolidated workflow here:

https://github.com/kingdonb/website/commit/dbd7c1aff076e754e491099a291115bc9cf5a869

It's not a full test, but it verifies that the new path-based trigger works, which has been the only part I cannot test locally.

makkes (approved):

kingdonb:
Thanks! The build for the base image succeeded

https://hub.docker.com/r/fluxcd/website/tags
^ shows we have multiarch images for the hugo base now, will have them for hugo-support next 🚀

PR #1585, opened 2023-07-27
docs: Use DoesNotExist operator in shard selector

jimmidyson:
This commit updates the docs to use `DoesNotExist` operator (`!`)
for sharding selector.

The current docs use `NotIn` (`notin`) operator, which means that each
time a new shard is added, the root flux deployments needs to be updated
to exclude the new shard label. By using `DoesNotExist` operator, the
root flux deployments can remain unchanged when a new shard is added.
This makes it slightly easier to add and remove shards.


makkes (approved):
Smart! Thanks, @jimmidyson

PR #1592, opened 2023-08-01
Prevent robots from indexing branch builds

makkes:
We don't want archived docs for old Flux versions to be indexed by search engines so that all pages at https://v2-0.docs.fluxcd.io don't turn up in search results.

All Netlify branch builds now lead to all pages being built and deployed to have the `noindex` directive in the page's header.

This PR can be merged immediately as it has no effect on how the site is build and deployed right now. It will only have an effect as soon as we start publishing branch builds for older Flux versions.

I will update the `v2-0` branch from `main` as soon as this PR is merged for the changes to take effect for that branch's deployment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ preview-build: gen-content prereqs ## Builds a preview build (for e.g. a pull re
 branch-build: gen-content prereqs ## Builds a Git branch (for e.g. development branches).
 	hugo \
 		--baseURL $(DEPLOY_PRIME_URL) \
-		--buildDrafts \
+		--environment branch \
 		--buildFuture \
 		--gc \
 		--minify \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -51,7 +51,7 @@ preview-build: gen-content prereqs ## Builds a preview build (for e.g. a pull re
 branch-build: gen-content prereqs ## Builds a Git branch (for e.g. development branches).
 	hugo \
 		--baseURL $(DEPLOY_PRIME_URL) \
-		--buildDrafts \
+		--environment branch
```
stefanprodan, line 54:
```suggestion
		--environment branch \
```

stefanprodan: We don’t need `buildDrafts` anymore?

makkes: I assume we don't need them at this moment. When you deploy a branch you don't need to mark anything as draft because it's not deployed at fluxcd.io, anyway.

stefanprodan (approved):

PR #1596, opened 2023-08-02
Add ZEIT ONLINE to adopters

scheying:
Thanks for the great work! FluxCD has become a natural part of our deployment ecosystem.


stefanprodan (approved):
Awesome! Thanks @scheying

PR #1600, opened 2023-08-02
Add Helm drift to config docs

mehak151:


stefanprodan (approved):
LGTM

Thanks @mehak151

PR #1601, opened 2023-08-02
Make import-flux2-assets.sh aware of the desired Flux version

makkes:
Since the website will be built for the 2.0.x minor release, the import script needs to fetch the API docs for the controller versions of the latest 2.0.x Flux release. This change makes the script aware of the desired version (derived from the `BRANCH` environment variable that is set by Netlify).

This change is a subset of the changes implemented in #1577. We cannot simply backport 1577 since the changes in there required some changes to the controller API docs which haven't been released as part of a Flux 2.0.x release, yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ static/install.sh
 
 # local GH token, used for some scripts
 .gh-token
+
+# build directories
+/bin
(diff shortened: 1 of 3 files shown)


makkes:
@stefanprodan this also needs to be merged before a 2.1 release so we don't generate the wrong docs on the 2.0 website.

PR #1606, opened 2023-08-03
Add version dropdown and deprecation notice

makkes:
I took the partial that's part of this change from the Docsy sources and slightly modified it to have the "deprecation-warning" class and some additional text. The original can be found here: https://github.com/google/docsy/blob/3f3172a190ac9b24aa5b9c695f62c3fee8273ac2/layouts/partials/version-banner.html

This commit also moves the link to Flagger into the top navbar so that we can actually use the version drop-down for listing Flux versions. A dedicated "Documentation" link in the top navbar now directs users to the Flux documentation.

Parts of this change will have to be forward-ported to `main` once 2.1 has been released. I'll create a PR for this later.

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -95,15 +95,16 @@ params:
   # Enable Algolia DocSearch
   algolia_docsearch: true
   offlineSearch: false
-  version_menu: "Docs"
-  version: "Flux"
-  url_latest_version: /flux/
+  version_menu: "Versions"
+  version: "2.0"
+  archived_version: true
+  version_menu_pagelinks: true
+  url_latest_version: https://fluxcd.io/flux/
   versions:
-    # Add your release versions here
-    - version: Flux
-      url: /flux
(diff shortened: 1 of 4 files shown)


makkes:
I need to figure out how to put the deprecation notice on the front page, too.

stefanprodan:
Is the search index set for each subdomain?

makkes:
Search on v2-0.docs.fluxcd.io actually searches fluxcd.io at the moment. As I see it, we have two options here:

- Disable search completely. This would not lead to confusion for users search in archived docs but ending up on results from the latest version without noticing and trying out configurations for a Flux version that might not work with the Flux version they're running. Drawback is that they would have to explicitly go to the latest version's site to search for anything.
- Leave it as it is. This way, you can always search for content, no matter if you're looking at an archived site or not.

stefanprodan (approved):
LGTM

Thanks @makkes 🏅

PR #1609, opened 2023-08-04
Add Disable k8 cluster role aggregations section in the config folder

mehak151:


stefanprodan:
This is already in here https://deploy-preview-1573--fluxcd.netlify.app/flux/installation/configuration/multitenancy/

I think we are done with the Configuration section, what we next it change all the links in websites that used to point to installation.md and bootstrap.md.

PR #1611, opened 2023-08-08
Fixed doc links for installation and bootstrap

mehak151:


content/en/flux/installation/configuration/air-gapped.md
```diff
@@ -63,4 +63,4 @@ flux bootstrap git \
 as the deploy key on your Git server in advance.
 
 For more information on how to use the `flux bootstrap git` command,
-please see the generic Git server [documentation](flux/installation/generic-git-server/).
+please see the generic Git server [documentation](../../bootstrap/generic-git-server/).
```
stefanprodan, line 66: I've used absolute links in the new docs, please change it here as well e.g. `/flux/installation/bootstrap/generic-git-server.md`

mehak151: Updated

kingdonb:
I think this fixes everything I would have missed in:

* #1621 

I'm approving, then merging it back to the install-v2 branch so the machine can check for any more broken links we've left behind. That should be (nearly?) all of them, though! 🥇 

(With the exception of those that will need to merge through another repo, tracked below)

* https://github.com/fluxcd/source-controller/pull/1200

kingdonb (approved):
Thanks for all your help with this @mehak151 and @stefanprodan

PR #1612, opened 2023-08-08
Add version dropdown and deprecation notice

makkes:
This is a forward-port of #1606 and #1608 in preparation of the 2.1 release.

I took the partial that's part of this change from the Docsy sources and slightly modified it to have the "deprecation-warning" class and some additional text. The original can be found here: https://github.com/google/docsy/blob/3f3172a190ac9b24aa5b9c695f62c3fee8273ac2/layouts/partials/version-banner.html

This commit also moves the link to Flagger into the top navbar so that we can actually use the version drop-down for listing Flux versions. A dedicated "Documentation" link in the top navbar now directs users to the Flux documentation.

The link to GitHub is also removed to make the navbar be closer to the one at kubernetes.io.

### DIFF
--- a/layouts/partials/navbar-version-selector.html
+++ b/layouts/partials/navbar-version-selector.html
@@ -1,0 +1,14 @@
+<a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown"
+	aria-haspopup="true" aria-expanded="false">
+	{{ .Site.Params.version_menu }}
+</a>
+<div class="dropdown-menu" aria-labelledby="navbarDropdownMenuLink">
+	{{ $path := "" }}
+	{{ if .Site.Params.version_menu_pagelinks }}
+	{{ $path = .Page.RelPermalink }}
+	{{ end }}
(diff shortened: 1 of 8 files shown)


makkes:
@stefanprodan this goes into the branch from which #1577 is created but I figured it's easier to review in batches.

layouts/partials/navbar-version-selector.html
```diff
@@ -0,0 +1,13 @@
...
+<div class="dropdown-menu" aria-labelledby="navbarDropdownMenuLink">
+	{{ $path := "" }}
+	{{ if .Site.Params.version_menu_pagelinks }}
+		{{ $path = .Page.RelPermalink }}
+	{{ end }}
+    <a class="dropdown-item" href="/flux/releases">Release Information</a>
+	{{ range .Site.Params.versions }}
+	<a class="dropdown-item" href="{{ .url }}{{ $path }}">{{ .version }}</a>
+	{{ end }}
+</div>
```
stefanprodan, line 13: Add a new line here please

stefanprodan (approved):

PR #1637, opened 2023-08-26
Update index.md

abohmeed:
Cosign needs the `sha256:` prefix for signing the chart.


kingdonb:
Thanks for pointing this out @abohmeed - could you please rebase and squash your commits? Also the DCO check failed on the last of them. Please take another look so we can merge it!

Thank you

PR #1649, opened 2023-09-04
fix Dockerfile for hugo-support image

kingdonb:
This should solve:

* https://github.com/fluxcd/website/actions/runs/6063171310

I am testing it out locally to be sure nothing else did change

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,6 @@ RUN apk update && \
 
 COPY requirements.txt /tmp
 RUN python3 -m pip install -r /tmp/requirements.txt
-RUN ln -s `which python3` /usr/bin/python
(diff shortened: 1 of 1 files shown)


kingdonb:
Looks as if somewhere in the distro or upstream rootfs from hugo is creating the symlink from `/usr/bin/python` to `python3` now, so there's no need to do it in the support image anymore. We could also update our `ln -s` to `ln -sf` (in case they later decide to put it back and undo what they did, but I doubt that will happen.)

kingdonb:
Once this is merged, the push-hugo-support job will run itself again automatically because it updates the Dockerfile 👍

stefanprodan (approved):
LGTM

Thanks @kingdonb

Issue #1658, opened 2023-09-14
Link not working


apuetz-cariad:
The link to the bootstrap cheatsheet here https://fluxcd.io/flux/components/helm/options/#:~:text=bootstrap%20cheatsheet is not working (page not found)

kingdonb:
Thanks for reporting this. We designed a link checker for the docs migration, but it didn't catch this 404.

Will find a way to make sure it's caught from now, and figure out what other broken links we have missed. 😭

makkes:
Thanks @apuetz-cariad for reporting this issue which is fixed now. 🎉

PR #1659, opened 2023-09-14
Fix link to bootstrap customization guide

makkes:
The re-organization that happened as part of https://github.com/fluxcd/website/pull/1573 removed this page.

PR https://github.com/fluxcd/website/pull/1573 added a redirect for /flux/cheatsheets/bootstrap but in this case it's better to specifically link directly to the customization guide instead of the general configuration page.

closes https://github.com/fluxcd/website/issues/1658


stefanprodan (approved):
LGTM

Thanks @makkes

PR #1664, opened 2023-09-20
Add a note for image automation to upgrade guide

stefanprodan:
Add example on how to upgrade when the Flux image automation feature has been enabled at bootstrap.

Closes: https://github.com/fluxcd/flux2/issues/4259

### DIFF
--- a/content/en/flux/installation/upgrade.md
+++ b/content/en/flux/installation/upgrade.md
@@ -43,6 +43,16 @@ git add -A && git commit -m "Update $(flux -v) on my-cluster"
 git push
 ```
 
+If you've enabled extra Flux components at bootstrap,
+like those required for the [image automation feature](/flux/guides/image-update/).
+Ensure to include these components when generating the components manifest:
+
+```shell
+flux install \
+--components-extra image-reflector-controller,image-automation-controller \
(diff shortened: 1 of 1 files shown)


mrwormhole:
thanks :+1:

content/en/flux/installation/upgrade.md
```diff
@@ -43,6 +43,15 @@ git add -A && git commit -m "Update $(flux -v) on my-cluster"
 git push
 ```
 
+If you've enabled the Flux [image automation feature](/flux/guides/image-update/) at bootstrap,
+make sure to include the image automation controllers when generating the components manifest:
```
hiddeco, line 47:
```suggestion
If you've enabled extra Flux components at bootstrap, like those required for the [image automation feature](/flux/guides/image-update/).
Ensure to include these components when generating the components manifest:
```

Makes it more future safe, in case we ever introduce new components.

In addition, maybe put this in some kind of warning / note / whatever block?

stefanprodan: There is an info box right after this, too many of those I think it just creates noise.

hiddeco (approved):

PR #1666, opened 2023-09-21
Update webhook-receivers.md

dudo:
Clear up a typo. Add a reference to the flux-system.

### DIFF
--- a/content/en/flux/guides/webhook-receivers.md
+++ b/content/en/flux/guides/webhook-receivers.md
@@ -58,7 +58,7 @@ Wait for Kubernetes to assign a public address with:
 watch kubectl -n flux-system get svc/receiver
 ``` 
 
-...or, create an `Ingress` with the same destination, the `notification-webhook` http service on port 80:
+...or, create an `Ingress` with a similar destination, the `webhook-receiver` service on port 80:
 
 ```yaml
 apiVersion: networking.k8s.io/v1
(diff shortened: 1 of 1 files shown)


kingdonb:
Confirming that `notification-webhook` is the name used for some reason in the doc, and `webhook-receiver` is the name used by Flux to expose the webhook receiver. 

```
webhook-receiver                                        ClusterIP   10.97.49.10      <none>        80/TCP     21d
```

All of my clusters with Flux installed have a `webhook-receiver` service that listens on port 80. I don't see any `notification-webhook` listed, only a `notification-controller` service. This must be a typo.

I'm not sure about adding the flux-system here. It makes sense to show `flux-system` in an example, I agree, but adding a second `GitRepository` has potential to confuse as well, since webhook would typically only trigger a single git repository source. What do you think @dudo?

content/en/flux/guides/webhook-receivers.md
```diff
@@ -132,6 +132,8 @@ spec:
   secretRef:
     name: webhook-token
   resources:
+    - kind: GitRepository
+      name: flux-system
```
stefanprodan, line 136: Please remove this. Adding flux-system which is a diffrent repo to the webapp will make Flux do the wrong thing.

stefanprodan (changes requested):

Issue #1667, opened 2023-09-22
Kustomization - Generating a kustomization.yaml file


johanneskastl:
(I wanted to open a PR for this page, but did not find it in https://github.com/fluxcd/website).

I really would like to see two important facts in this page. And in the quickstart guide, but I'll open another issue for that.

- if no kustomization.yaml (Kubernetes, not the Flux2 one) is present, **all** manifests from **all** subdirectories are used.
- you can exclude manifests by using a kustomization.yaml that does not contain the manifests you want to exclude.

Something like this:

```
By default, Flux uses all manifests in the path specified by the Kustomization, if no kustomization.yaml is present. In this case a kustomization.yaml will be generated on the fly using all the manifests found in this directory and its subdirectories.

If a kustomization.yaml is found, only those manifests mentioned in it will be used. This is useful to exclude manifests or folders from being used, to e.g. use them in a separate kustomization due to ordering issues
```

PR #1668, opened 2023-09-22
guides/repository-structure.md: add information on kustomization.yaml and ignoring/excluding files

johanneskastl:
Add a preface regarding the presence of a kustomization.yaml and the consequences, i.e. that files can be excluded/ignored by using a kustomization.yaml

I would also like to see this information in the Kustomization component page, see #1667


stefanprodan:
I don't think this techical detail should be in here. This doc is more high level, maybe this can be at the bottom with a link to the https://fluxcd.io/flux/components/kustomize/kustomizations/#generating-a-kustomizationyaml-file instead of duplicating the same info.

stefanprodan:
Maybe we could have an info box that points at https://fluxcd.io/flux/faq/#can-i-use-repositories-with-plain-yamls

johanneskastl:
> Maybe we could have an info box that points at https://fluxcd.io/flux/faq/#can-i-use-repositories-with-plain-yamls

That would be helpful, but also only tells have of the story. My main point is that be default the auto-generated kustomization.yaml contains all manifests, while you can select and exclude if you create one yourself.

stefanprodan:
You can exclude files using GitRepository `spec.ignore`, what's written in https://fluxcd.io/flux/faq/#can-i-use-repositories-with-plain-yamls

johanneskastl:
> You can exclude files using GitRepository `spec.ignore`, what's written in https://fluxcd.io/flux/faq/#can-i-use-repositories-with-plain-yamls

Yes, I also found that. But I personally find the kustomization.yaml approach more appealing, because it means I find the root cause for a file exclusion where the excluded files are residing. Specifying that in the GitRepository means it is stored in another place, where it might not be obvious to look.

In the end I think it would not harm to have an infobox or similar on the page, just to make newcomers aware of this. There is more than one way to use Flux and this might be helpful for some.

johanneskastl:
@stefanprodan How to proceed? I would really like to see some mention of this in the docs, but of course it is up to you to decide which way you want it to be...

PR #1681, opened 2023-09-29
FAQ: Explain how to use the `prune` annotation when moving resources

stefanprodan:
Add alternative procedure to `How can I safely move resources from one dir to another?`


kingdonb (approved):
LGTM

PR #1709, opened 2023-11-08
Add Tchibo to flagger adopters

lacunoc:
Tchibo wants to provide its developers progressive delivery tooling for its internal developer platform. As Flux is already in use, Flagger perfectly complements the GitOps approach here.
An example application implementation, which covers several use cases, got already provided for internal usage for our developers.


stefanprodan (approved):
Thank you @lacunoc 🥇

PR #1716, opened 2023-11-13
Fix bad link in get-started.md

kingdonb:
The disambiguation page is "Bootstrap" now

We can merge this straight away, or wait until I've reviewed all the links on this page. I think we're trying to avoid relative links now generally, so I'll change all of the relative links on this page into absolute paths even if they're not currently bad links.


kingdonb:
I think this is good to merge. Ready for review

stefanprodan (approved):

PR #1720, opened 2023-11-17
Add in-memory kustomize builds advice to scaling docs

stefanprodan:
Recommend using tmpfs based on benchmark results from https://github.com/stefanprodan/flux-benchmark 

Part of: https://github.com/fluxcd/flux2/issues/4410


makkes (commented):
We should add a note that using tmpfs will count against the pod's memory usage. I wouldn't expect this to have a huge impact in most of the cases but depending on the number of manifests it might pose an issue.

stefanprodan:
Preview here: deploy-preview-1720--fluxcd.netlify.app/flux/installation/configuration/vertical-scaling/

makkes (approved):
🎉

Issue #760, opened 2022-02-11
Refer to "core concepts" earlier and more


dholbach:
https://fluxcd.io/docs/concepts/ explains core concepts like "Kustomization". This is great.

We should re-work the flow of our docs a little, so new users are directed to this more early on, e.g. on `/docs/` or `/docs/get-started` where e.g. Kustomization is mentioned and it's sort of assumed that you know what it and its impact is - it will make the life unsuspecting and beginner type of readers easier.

kingdonb:
Tracking in #1723 from now. 📇

PR #1723, opened 2023-11-20
Added more terms to Core Concepts

mehak151:
As per https://github.com/fluxcd/website/issues/760, the Core Concepts page is already positioned at the top of the docs, before `Get Started`. 

This PR aims to ensure the user has only 1 glossary to refer to, instead of 2 on different pages, and they get to learn about all the basic terminologies.

### DIFF
--- a/content/en/flux/concepts.md
+++ b/content/en/flux/concepts.md
@@ -99,3 +99,36 @@ Progressive Delivery builds on Continuous Delivery by gradually rolling out new 
 Developers can use techniques like feature flags, [canary releases](https://glossary.cncf.io/canary-deployment/), and A/B testing to minimize the chances of introducing bugs or errors that could harm users or interrupt business operations. These strategies enable a controlled and gradual rollout of new features, ensuring a smooth and successful release that enhances user trust and improves the overall user experience.
 
 The Flux project offers a specialised controller called [Flagger](https://github.com/fluxcd/flagger) that implements various progressive delivery techniques. For more information, take a look at [Flagger deployment strategies](https://fluxcd.io/flagger/usage/deployment-strategies/).
+
+## Cluster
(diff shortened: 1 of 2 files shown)


mehak151:
> I know I suggested this change originally, but I am afraid the glossary and the core concepts are not the same. There are some cross purposes at play here, the original intent of the e2e page was to create a single page where you could go to read everything about how Flux works all at once. I think this is the right change for the e2e page, but I'm not sure it's right to go there.
> 
> Also I feel compelled to reorganize these terms and put them in alphabetical order, but the core concepts page is not in alphabetical order.
> 
> https://deploy-preview-1723--fluxcd.netlify.app/flux/concepts/#service
> 
> There's also a typo in the translation here, one of the headings has `**` still appended to it

@kingdonb I can go ahead and make the following changes if we are good to go with it:

1. Rearrange glossary in alphabetical order
2. Remove the **
3. Fix the typo

However, before I do that, I want to share that I understand why the glossary was kept separate from the core concepts. Having that said, don't you think all the terminologies should be discussed under the same page instead of maintaining two different pages? What are your thoughts?

kingdonb:
I think we should have one page for terms. I'm good to merge this and iterate once we fix the typo 👍

kingdonb:
Sorry, I brought up alphabetical order because the glossary was in alphabetic order, but the Core Concepts is not a glossary, it begins with the most important concept and proceeds vaguely chronologically through Kustomization which is the end.

I don't know if the glossary adding to the core concepts is going to make things too populated too early. One of the features of the Core Concepts page currently is it does not have very many ideas on it, only the ones that are most central to Flux.

stefanprodan:
Note of the added things are core concepts of Flux.

PR #1754, opened 2023-12-15
Polish Gitea PAT permission statements

makkes:
The doc now explicitly mentions the required minimal permissions.


stefanprodan (approved):
LGTM

Thanks @makkes

PR #1779, opened 2024-01-19
Added SSH over HTTPS port warning to github bootstrapping docs

dlactin:
Added a warning for an unlikely but potential issue for folks who are using SSH over port 443 when bootstrapping flux. 

I ran into this issue and even with the host keys set to the ones in that link I was unable to get this to work until I changed the port back to 22.

### DIFF
--- a/content/en/flux/installation/bootstrap/github.md
+++ b/content/en/flux/installation/bootstrap/github.md
@@ -159,5 +159,11 @@ flux bootstrap git \
 
 **Note** that you must generate a SSH private key and set the public key as a deploy key on GitHub in advance.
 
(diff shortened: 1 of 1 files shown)


dlactin:
Hello! 

Not a problem. 

It's been a while since I've looked at this, but from what I recall, your second comment is spot on. 

> Are the affected users Enterprise GitHub customers who have set up SSH on port 443, for some firewall reason?

This was implemented while a few of us were at a conference that had port 22 blocked, and it caused our attempts at bootstrapping a cluster with flux to fail without a generic host key mismatch error. 

Definitely not a flux problem, but something I thought would be worth mentioning in the docs for folks bootstrapping flux this way.

PR #1781, opened 2024-01-21
Fix Hugo 0.112.0 deprecation

kingdonb:
https://gohugo.io/content-management/multilingual/#changes-in-hugo-01120

While cruising the logs I noticed a deprecation warning that can be fixed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ preview-build: gen-content prereqs ## Builds a preview build (for e.g. a pull re
 branch-build: gen-content prereqs ## Builds a Git branch (for e.g. development branches).
 	hugo \
 		--baseURL $(DEPLOY_PRIME_URL) \
-		--environment branch \
+		--environment=branch \
 		--buildFuture \
 		--gc \
 		--minify \
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -71,10 +71,11 @@ markup:
 
 languages:
   en:
-    languageName: English
-    languageCode: en
-    time_format_default: 2006-01-02
-    time_format_blog: 2006-01-02
+    params:
+      languageName: English
+      languageCode: en
+      time_format_default: 2006-01-02
+      time_format_blog: 2006-01-02
 
 permalinks:
   blog: /:section/:year/:month/:slug/
--- a/netlify.toml
+++ b/netlify.toml
@@ -4,7 +4,7 @@
 
 [build.environment]
   HUGO_VERSION = "0.121.2"
-  NODE_VERSION = "16"
+  NODE_VERSION = "18"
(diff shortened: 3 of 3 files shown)


kingdonb:
Without [5bff4d1](https://github.com/fluxcd/website/pull/1781/commits/5bff4d1b7f750a457b9991665d98c4b24893679e), on MacOS the `make branch-build` is failing with an error unknown flag: branch

I tested the change locally and it worked. I am not sure why this doesn't seem to fail on the netlify builds. Also weird that when I log into Netlify with my github credential, it blocks me from viewing the build logs, but if I remain unauthenticated I can see they are public (they're just trying to get me to log in by hiding all but the first few... not very well hidden either)

If we merge this PR, it will no longer be possible to tell if the calendar updates, since merging a PR will change some file and then it will definitely update. I'm not any closer to understanding why netlify doesn't see the calendar changing as an update. If there is some script that short-circuits out of running because the data source hasn't changed, or what... I think I need to monitor this over a couple of days.

We can merge this even if it makes debugging a little harder, because the main problem is the calendar isn't updating, and merging it will update the calendar to show the next dev meeting anyway. Root cause can wait until later, I don't think I should need any greater access to solve this than what's in public and what I already have.

kingdonb:
This is the page where I can see build logs, as long as I don't log in 😆 

https://app.netlify.com/sites/fluxcd/deploys

kingdonb:
I'll see if I can resolve all of the build warnings here

kingdonb:
I've just opened a rabbit-hole that I may wish to put back closed, it seems as if Netlify around the beginning of last year close sourced their `netlify/build-image` so they could "move faster" according to the README in the archived repo.

Since then, there have been numerous forum posts asking for a newer version of Python, but only Python 3.8 is supported, and as of late last year there was no update on this situation. Python 3.8 has been out of support since:

> Ended 2 years and 8 months ago
> (03 May 2021)

and security support will end within a year. We may wish to make an adjustment here. I don't really have any problem with Netlify in general, but while I was debugging the calendar thing I came across this issue.

(Reverting that part of the upgrade, we cannot upgrade Netlify from Python 3.8)

kingdonb:
Noting for the first time that a merged PR doesn't seem to advance the calendar... 1/23/24

kingdonb:
I'm removing the two commits that no-op upgrade/reverting the python, since that didn't work.

The rest of these fixes are trivial but necessary upgrades to resolve build warnings, simply so the next person doesn't have to look at them! 💐 

One of the fixed issues caused `make branch-build` to fail on MacOS at least. It didn't seem to fail in the netlify pipeline.

stefanprodan (approved):
LGTM

Thanks @kingdonb

Issue #1787, opened 2024-01-24
Website calendar does not advance anymore


kingdonb:
We figured out why this is happening, in a conversation that happened outside of GitHub.

The answer is here: https://fluxcd.io/flux/releases/procedure/#backport-changes-for-patch-releases

Because the every-6-hourly job only triggers the build on the `main` branch, nothing is updated on any other branch. That `main` branch is reflected somewhere, but it's not the main website that we advertise on https://fluxcd.io

The website workflow I'm talking about is here:

https://github.com/fluxcd/website/blob/main/.github/workflows/netlify.yml

You can see that it's scheduled to run every 6 hours, this is done so that the calendar will advance itself and articles that are scheduled to publish on a particular date will be able to automatically become published on that date. It just triggers a rebuild.

We can update this to trigger the `v2-2` branch to build instead, I have to look into it to see how exactly (the job referenced here builds but doesn't call out the `main` branch specifically, I'm sure there is a way that we can pass the branch as a parameter...)

~The only hangup that we identified is that `v2-2` is a moving target, and should advance every quarter. We could just write the latest minor version in the file and advance it whenever a new release comes, I don't think anyone is in favor of doing that. We don't need one more extra step in the release protocols, it's complicated enough! I will write the logic as a part of the bash script in `netlify.yml` and test outside the Flux repo...~ I didn't wind up doing all this, although I found https://gist.github.com/bitmvr/9ed42e1cc2aac799b123de9fdc59b016 which isn't so complicated as I imagined parsing semver in bash, we need a new netlify build hook trigger for each new branch to be built from, so it's going to be complicated to script this. More details in #1795

We can borrow some logic from another script to look up the latest released version of Flux, then do some string manipulation to generate the branch name that would be associated with the latest minor version.

I'm working all this out, I will solve it this week, just wanted to see that the issue and plan to resolve it is documented first.

PR #1791, opened 2024-01-28
[v2-2] update hugo to v0.122.0

kingdonb:
(Manual backport of #1790, because the label trigger noticed I added a merge commit, the workflow called me out in https://github.com/fluxcd/website/actions/runs/7688160859/job/20949053159#step:3:35)

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,7 +3,7 @@
   command = "make production-build"
 
 [build.environment]
-  HUGO_VERSION = "0.121.2"
+  HUGO_VERSION = "0.122.0"
   NODE_VERSION = "18"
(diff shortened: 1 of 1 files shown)


kingdonb:
I am still working on:

* #1787 

so we need to be sure that v2-2 gets a PR merged every so often, to keep the calendar advancing.

Otherwise, I do not think this would be a critical backport, but at this point I am not sure I have the sufficient permissions in Netlify to generate a build hook or to trigger the build hook for v2-2.

Merging this PR should advance the calendar, though...

stefanprodan (approved):

kingdonb:
Thanks @stefanprodan!

PR #1807, opened 2024-02-11
update podinfo GitRepository branch to match guide

seanknox:
The podinfo `GitRepository` resource was set to use `master` instead of `main`.


seanknox:
nm, misunderstanding on my part.

PR #1815, opened 2024-02-20
Blog post: Introducing Capacitor

laszlocph:
Hi Flux maintainers,

We have been working on a general purpose Flux GUI over at: https://github.com/gimlet-io/capacitor
The goal is to be the go-to dashboard for debugging applications that Flux deployed.

We are at the beginning of the journey, hoping to fill the void of Flux GUIs. It is Apache licensed and our plan is to keep it that way. Keep the tool un-opinionated and applicable on all clusters where Flux is used. With a hint of UI/UX opinions though - not a fan of tables of K8s resources.

This is a blog post proposal for your blog, let us know if this is something that you would publish! It would help early adoption.

Best regards,
Laszlo & Capacitor maintainers.


stefanprodan:
The CI failure is unrelated to the content, looks like the preview script don't work for PR from forks, @makkes is looking into it.

makkes:
This is a very strange issue with the failing deploy preview. I'm pretty sure Netlify is somehow using outdated code because e.g. it says

```
7:03:05 AM: Installing Hugo 0.115.1
```

even though the Hugo version in the `netlify.toml` file is set to 0.122.0.

Then it proceeds to use an outdated `import-flux2-assets.sh` script so the build fails. 🤔

laszlocph:
It produced the same result for me locally.

I commented out some codepaths to ignore it.. thought it is my hugo setup.. just want to say, maybe you can reproduce the issue on a clean install.

makkes:
> It produced the same result for me locally.
> 
> I commented out some codepaths to ignore it.. thought it is my hugo setup.. just want to say, maybe you can reproduce the issue on a clean install.

Thanks, that helps a lot. Let me check locally, too.

makkes:
Running

```
BRANCH=main ./hack/import-flux2-assets.sh
```

works fine locally so I suspect it's a Netlify issue somehow.

laszlocph:
Oh, this was surprising. Now rebased and all good. Sorry for the mess.

kingdonb:
Thanks for pushing this PR! I will take a look at Capacitor, at a glance the article seems to end with no conclusion, which may be OK since the call to action is "go, install capacitor" - so the second feedback I would offer is, did you consider adding signatures for your OCI repository artifacts with cosign?

You can do it in CI without managing any keys, in a way that people can verify your subject/issuer don't change:

https://fluxcd.io/flux/cheatsheets/oci-artifacts/#workflow-example

This is just my gut reaction upon looking at an OCIRepository example that does not include the `verify` field in spec. I don't think it's a blocker for publishing, but opinions may vary, I think it's an easy enhancement that would let us add a bit more content to round out the conclusion (and help us to educate more Flux users about the new-ish feature)

stefanprodan (approved):
LGTM

Thanks so much @laszlocph for creating Capacitor ❤️

stefanprodan:
@laszlocph I propose we go live with it on Monday. Also can you please open a new PR and add Capacitor to the ecosystem page with 2 screen side-by-side https://fluxcd.io/ecosystem/#flux-uis--guis

kingdonb:
I can still merge this to go out today, just checking since it did not go out on Monday as planned.

Is it ready to go?

laszlocph:
> Is it ready to go?

We are ready on our side!

kingdonb (approved):

PR #1826, opened 2024-02-25
Fix aenix logo in adopters

kingdonb:
Noticed the logo isn't showing up in Adopters, it isn't because of any reason except for mismatched filename.

I have checked that SVG logos are OK here and it appears they are. (@kvaps FYI thanks for adding this here!)


kvaps (approved):
LGTM, thanks for fixing this ;)

PR #1838, opened 2024-03-03
Add image policy example for DockerHub format

stefanprodan:
Add example on how to select the latest version by extracting the semver from tags in the format `<semver>-<distro>`
e.g. [Redis on DockerHub](https://hub.docker.com/_/redis).


kingdonb (approved):
👍 🥇

PR #1854, opened 2024-03-07
Remove ResourceQuota from sharding overlay example

relu:
The ResourceQuota resource should not be duplicated by the sharding kustomize overlay.


stefanprodan (approved):
LGTM

Great catch! we added the `ResourceQuota` later on and we didn't updated the patch.

PR #1882, opened 2024-03-28
Added KCL controller to list of extensions

kim-sondrup:


kingdonb (approved):
Thanks for contributing this @kim-sondrup 🎉 

I'm not super familiar with KCL, but from a brief look at the links you provided I can tell it's something I should take a closer look at. It was definitely also something people brought up to me at least at KubeCon Paris, when we were at the Flux booth!

I think it fits great here on the flux ecosystem page.

PR #1893, opened 2024-04-04
add timestamp to fix calendar sort

kingdonb:
Fix issue introduced in #1891 

(the issue: the team calendar's sort order is now out of whack)

Screenshot 2024-04-04 at 8 30 45 AM

There was some implicit sorting and some re-sorting, it's probably clearer if we pass the original timestamp through to the "formatted_event" and sort by that field (even though it is not used directly for display)

### DIFF
--- a/hack/import-calendar.py
+++ b/hack/import-calendar.py
@@ -96,6 +96,7 @@ def read_calendar(cal):
                 'label': str(event['summary']),
                 'where': format_location_html(event_location),
                 'description': description,
+                'timestamp': event_time,
         }
 
         if 'organizer' in event:
@@ -104,7 +105,7 @@ def read_calendar(cal):
 
         events.append(formatted_event)
 
-    events.sort(key=lambda e: e['time'])
+    events.sort(key=lambda e: e['timestamp'])
     return events
 
 def format_location_html(location):


hack/import-calendar.py
```diff
@@ -96,6 +96,7 @@ def read_calendar(cal):
                 'label': str(event['summary']),
                 'where': format_location_html(event_location),
                 'description': description,
+                'timestamp': event_time,
```
makkes, line 99: `event_time` is not a timestamp but a `datetime.datetime`.

makkes (changes requested):

hack/import-calendar.py
kingdonb: The output in the YAML file is shaped like this (Python string formatter coerces it into a timestamp):

```
  timestamp: 2024-04-04 15:00:00+00:00
```

I've been planning to provide localized time with client-side JS, having the timestamp in a format that I can reliably parse back will be invaluable.

kingdonb: Yes, I think we can rely on the Python string formatter to always format DateTime in a way that I can pass to Time.parse in any language down the line. A unix timestamp would also suffice for my needs here.

makkes (approved):

hack/import-calendar.py
kingdonb: Thanks for your review! 🙇‍♂️

PR #1897, opened 2024-04-05
Fix parsing calendar with organizer

makkes:
When the ICS file contains an organizer, the script failed because it didn't parse the field correctly.


kingdonb (approved):
👍 LGTM

PR #1900, opened 2024-04-07
Fix calendar import

makkes:
The existing URL started returning 404s so I changed it to the one that it was apparently replaced with.

For posterity, you go to https://lists.cncf.io/g/cncf-flux-dev/calendar, scroll down, click on "Subscribe to Calendar" and copy the URL from the subsequent dialog.

### DIFF
--- a/hack/import-calendar.py
+++ b/hack/import-calendar.py
@@ -27,7 +27,7 @@ import recurring_ical_events
 import urllib3
 import yaml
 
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
Thanks @makkes

kingdonb (approved):
Thanks @makkes 💯

PR #1902, opened 2024-04-07
Upgrade go to 1.21, clean up build issues

kingdonb:
I have narrowed down the issues upgrading hugo to a docsy v0.7.0 release breaking change that brings Bootstrap 5.

I don't know the exact nature of the errors I get with those upgrades, but the good news is it looks like Hugo upstream is already using Go 1.21, so we should at least upgrade to that version to match.

I will investigate the Bootstrap 5 upgrade situation further and get back to us soon about it. In the mean time, I've done loads of testing with `make production-build URL=https://fluxcd.io` on different versions of go modules and go compilers, to determine that this combination works, and we do not have to stay on Go 1.19 any longer.

(Doing one more test locally, since I do not think there is any CI workflow other than the hugo-support image builder that will trigger, and that triggers on merge, so there will be no other chance or automated way to test this before merge)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ARG HUGO_VERSION
 FROM fluxcd/website:hugo-${HUGO_VERSION}-extended
-COPY --from=golang:1.19-alpine /usr/local/go/ /usr/local/go/
+COPY --from=golang:1.21-alpine /usr/local/go/ /usr/local/go/
 
 ENV PATH="/usr/local/go/bin:${PATH}"
 
--- a/content/en/.gitignore
+++ b/content/en/.gitignore
@@ -2,4 +2,5 @@
 community.md
 contributing.md
 governance.md
+kubecon.md
 security.md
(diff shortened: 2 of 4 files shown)


kingdonb:
Link to the docsy changelog for v0.7.0 (where BREAKING CHANGES have been indicated):

https://github.com/google/docsy/releases/tag/v0.7.0

kingdonb:
What I have intentionally not done in terms of upgrading inside of this PR is:

- [ ] `hugo mod get`

This skips upgrading Docsy past `v0.7.0` to `v0.9.0`, we stay pinned at `v0.6.0` but we're catching up to the second-to-latest Go and removing a warning that would be emitted during `make production-build`.

I do not know how `SearchIndex` works, so cannot say if it was/is broken before or after the upgrade (but the message is properly silenced by the update to `hugo.yaml` `outputs`) - these things all seem to be safe to merge!

stefanprodan (approved):
LGTM

Thanks @kingdonb 🎖️

Issue #1899, opened 2024-04-06
Make use of `flux bootstrap --registry-creds` in the air-gapped installation guide


stefanprodan:
Followup: https://github.com/fluxcd/flux2/pull/4706

PR #1903, opened 2024-04-08
Improve the air-gapped with `flux bootstrap --registry-creds`

stefanprodan:
Followup: https://github.com/fluxcd/flux2/pull/4706
Ref: https://github.com/fluxcd/flux2/issues/4621
Fix: #1899

PS. This PR can be merged into main and must be published under `v2-3` branch after Flux v2.3 release.

### DIFF
--- a/content/en/flux/installation/configuration/air-gapped.md
+++ b/content/en/flux/installation/configuration/air-gapped.md
@@ -15,32 +15,26 @@ download the Flux CLI from [GitHub releases page](https://github.com/fluxcd/flux
 
 List the Flux container images with:
 
-```console
-$ flux install --export | grep ghcr.io
-image: ghcr.io/fluxcd/source-controller:v1.0.0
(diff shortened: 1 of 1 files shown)


content/en/flux/installation/configuration/air-gapped.md
```diff
@@ -64,3 +57,33 @@ as the deploy key on your Git server in advance.
 
 For more information on how to use the `flux bootstrap git` command,
 please see the generic Git server [documentation](/flux/installation/bootstrap/generic-git-server/).
+
+## Configure the image pull secret
```
souleb, line 61: Can we name this section something like `Boostrap Flux with a private registry`? This is still about bootrapping, even though there is an extra config strp.

stefanprodan: In air-gapped the registry is always private.

souleb (approved):

PR #1917, opened 2024-04-15
Localize meeting times in the browser with JS

kingdonb:
The javascript detects the browser's local time zone and converts everything formatted as a calendar row (or it just does nothing when javascript is not enabled) - the timestamp is hidden using `aria-hidden` and CSS: `display: none`.

The "Stimulus" package was invaluable to make this an easy update. I used data-target and data-controller properties to tell the JS code which elements to act on. It should be possible to follow what the code is doing, it's not very complicated.

Check out the Deploy Preview here:
* https://deploy-preview-1917--fluxcd.netlify.app/

The "next-event" section should also be localized by this PR, so if you're looking for any meeting time it gets localized.

In an earlier revision of the spec for this idea, the user was intended to be able to choose their time zone on the page. I have not implemented the time zone picker and I don't think it's needed anymore.

This PR also adds a "spin up" scaffold so you can type `make spin-up` and serve the `public/` directory, for a reasonable local test. This only runs `hugo build` that takes about 4 seconds so it's pretty good for iterative development of CSS and JS. The JavaScript is actually built by hugo, in spite of what I thought, so any syntax errors should be caught at build time.


kingdonb:
Really pleased with how this comes out.

Screenshot 2024-04-15 at 10 57 36 AM

There's definitely a bit more simplification we can do inside of the code, but this works!

kingdonb:
I did some normalization to en-UK locale because too much locale variability will make it hard to predict how much space is needed for formatting the team calendar. I think everyone can understand 24 hour time fine if it's in their local time zone.

I also did some testing for different time zones because the way of printing the local time zone varies a bit from zone to zone. Some places want to say something like "CEST" (or MESZ localized) and some places want to say "GMT+X" - I have used the en-UK locale to flatten this. And included the offset to emphasize for the reader that these time zones **are in fact localized** for them, though this may make it less obvious when JS is disabled that the time zone is not localized anymore.

Tested across some "more difficult time zones" with half-offsets like Calcutta, Moscow, Shanghai, Tokyo, they all should be fine and the formatting should be fine. Mainly testing to see nothing is too wide making the row spill onto more than one line.

(LMK if you see one, especially your local time, that isn't formatted just right!)

I would appreciate it if some different people can take a look at the preview deploy and verify that the JavaScript localization actually worked, that it picked your correct time zone and correctly applied it in both the next-event and the team calendar:

Screenshot 2024-04-15 at 12 36 08 PM

Screenshot 2024-04-15 at 12 36 19 PM

Added bonus should be when the time zone changes you can actually see it changing, as each localized time is calculated with the local time based on the date used. So if the time changes to GMT-5 tomorrow, you'll see that change on that date.

kingdonb:
One last update. I think the team calendar layout has been made OK for any width. It actually works a bit better now if we use the user's locale to decide how to format the time. So if you're in US, you get 12 hour time with AM/PM, "EDT, PDT" and you get M-D-YYYY dates like US people expect. If you're in UK, anywhere in Europe, you get 24 hour time with DD-MM-YYYY.

If you are in Mumbai you should get Hindi numerals (which may be perfectly OK and what you expected as well): १७-४-२०२४

(Let someone from any of these regions chime in and tell us if these are not formatted as you'd expect them, please!)

kingdonb:
Removed one last bit of nonsense where I was reformatting slash-separated dates as dash-separated. Left it the way it was in favor of honoring the locale, since there are some locales that use dashes, some use slashes, and some use periods!

I have expanded my horizons a bit, tested this against every locale I can think of, and invented some more. I thought we might add the previously talked about dropdown box to select the time zone, so that the visitor gets (in addition to their local time zone which gets used by default) some actual Flux developer time zones, IDK, maybe nice to have.

One other thing that might make for a nice enhancement is using the `Intersection` behavior from StimulusUse:

https://github.com/stimulus-use/stimulus-use/blob/main/docs/use-intersection.md#usage

So rather than all the JavaScript events firing on page load, they all fire when the HTML element they effect is exposed in the viewport. We would just move some code from the "connect" method into the "appear" function, and that code only runs when the element appears. As it defers some operations until later, it could wind up making some page loads a bit faster.

I did remove the additional StimulusUse library because I thought, we should add only one library at a time, we don't need this additional behaviors library, but if there are any performance concerns, I think it would be better to tackle them 🙃 the StimulusUse library includes other nice things like debouncing and throttling, that we don't want to be writing from scratch.

stefanprodan:
Do we have an Australia Edition for bug scrub?
Screenshot 2024-04-16 at 20 02 24

kingdonb:
Yes, @xUnholy has been hosting it alternating with me, this meeting is coming up in just under 4 hours from now!

stefanprodan (approved):
LGTM

Great UX improvement! Thanks @kingdonb 🥇

PR #1921, opened 2024-04-17
Update OpenShift bootstrap instruction

stefanprodan:
Followup: https://github.com/fluxcd/flux2/pull/4729
Fix: https://github.com/fluxcd/flux2/issues/4609


souleb (approved):
LGTM!

PR #1940, opened 2024-04-24
Move some of calendar to community.md page

kingdonb:
NOTE: Before this merges:

- [x] Rebase out or revert 294d423 - this is necessary to make preview-build work here,
        until fluxcd/community#396 merges

Check out the Deploy Preview here: https://deploy-preview-1940--fluxcd.netlify.app/
(note changes:
* Our Team Calendar: https://deploy-preview-1940--fluxcd.netlify.app/#calendar
* Community: https://deploy-preview-1940--fluxcd.netlify.app/community/#getting-involved)

These are the changes we discussed at the Flux Dev Meeting yesterday.

PR notes:

Let's reduce the amount of text and noise on the landing page, we only need to show the first few events.

In order to use the shortcode for the calendar in the external-imports, we need to avoid having parts of its HTML content converted into `<pre>` sections by the hugo builder.

It's an annoying change, but this is the only external-import with any shortcode that I'm aware of, and shortcodes aren't very big, so I think we can do without the indenting 🤷 

Some text changes and display formatting. The CSS rules to put the calendar into dark mode have been cloned and revised so the look of the table on the Community page is at least halfway decent. The calendar on the main page is now rendered by a `calendar_dark` shortcode, instead of the old `calendar` shortcode, but both are kept as I do not want to remove the calendar from the landing page yet. Just reduce its footprint.

(The preview has been made to work, even while fluxcd/community#396 hasn't merged)

### DIFF
--- a/assets/js/calendar.js
+++ b/assets/js/calendar.js
@@ -14,7 +14,7 @@ $('.calendar-list li').click(function() {
     static targets = [ "localtz" ];
     connect(){
       let tz_description = Intl.DateTimeFormat().resolvedOptions().timeZone;
-      this.localtzTarget.innerHTML = `<b>now localized for ${tz_description}</b>`;
+      this.localtzTarget.innerHTML = `<b>localized for ${tz_description}</b>`;
     }
   })
 
--- a/assets/scss/blocks/_blocks.scss
+++ b/assets/scss/blocks/_blocks.scss
@@ -2,5 +2,6 @@
 @import "cncf";
 @import "hero";
 @import "calendar";
+@import "calendar_dark";
 @import "blog";
 @import "features";
--- a/assets/scss/blocks/_calendar_dark.scss
+++ b/assets/scss/blocks/_calendar_dark.scss
@@ -1,18 +1,18 @@
-.team-calendar {
-  padding: 0 0 1.5rem 1rem;
+.team-calendar-dark {
+  padding: 0 1rem 1.5rem 1rem;
   max-width: 740px;
-  // margin: 0 auto;
+  margin: 0 auto;
   @include media-breakpoint-down(sm) {
     overflow-x: hidden;
   }
 
   h3 {
     font-weight: 600;
-    text-align: left;
+    text-align: center;
   }
   p.caption, p.details {
-    text-align: left;
-    margin: 1rem 0 1rem 1rem;
+    text-align: center;
+    margin: 1rem 0 2rem 0;
   }
 
   p.details {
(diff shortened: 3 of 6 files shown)


kingdonb:
We cannot just merge either this PR or fluxcd/community#396 - they must both merge at the same time, that one first, then this one.

Actually, there is no way to preview this except locally, because the changes in fluxcd/community when merged will take effect immediately.

kingdonb:
The result of the indenting in `home/calendar` shortcode file which is resolved in 123f1ea:

Screenshot 2024-04-24 at 12 59 42 PM

It may be more complicated than I thought to put a shortcode in an external sources file. Removing the indenting gets us the next image:

Screenshot 2024-04-24 at 1 07 22 PM

Like I mentioned in the PR description, this doesn't look very good. It might warrant the creation of a totally new CSS style for this page. I'm not prepared to tackle that today, but here is a starting look at what this PR could be.

I can hack something together so we actually have a live preview site deployed with my new `flux-docs` scaffold, it will be somewhere else - not in the usual place where netlify previews go. Given this change breaks some formatting I am going to pull it back into a draft PR for now, it will take some time to get this right. Maybe even a little while after 2.3 is released.

kingdonb:
The deploy preview is at least showing the progress now:

https://deploy-preview-1940--fluxcd.netlify.app/community/#getting-involved

kingdonb:
Screenshot 2024-04-24 at 2 25 14 PM

On the Community page, without dark mode, the calendar can look as above with a few CSS changes I'll keep working on

The external-sources page presents a new set of challenges though:

Screenshot 2024-04-24 at 2 25 22 PM

These links used to be html escaped, but something is preventing that escape now and the HTML links don't make it through the rendering. That's going to be enough to stump me for at least the rest of the day. 🦮🚶 Time to take the dog for a walk.

kingdonb:
Not sure why the simple escape `| safeHTML` wasn't needed when the calendar was on the landing page!

It turns out that escaped link isn't such a big deal to fix. I've adjusted the CSS rules so things should at least look right on the landing page and on the community page simultaneously. There is a huge hunk of CSS that has just been copied and I don't want that to be merged, there are a few other things to do here before it can be seriously considered for merging.

I've split the calendar shortcode into two shortcodes "calendar" and "calendar_dark" so we can keep some shade remnant of the calendar on the landing page, (so we don't break all of those references to fluxcd.io/#calendar out there in public)

I think it makes sense to show the next 5 or 10 days of events, but once we have the calendar in place on the community page and everything is fine, we may remove that calendar entirely from the landing (and replace it with a single sentence -> link to the community page)

kingdonb:
Looking a lot better now:

Screenshot 2024-04-24 at 5 49 03 PM

and the `safeHTML` invocation works absolutely fine (so long as you do remember to actually call it):

Screenshot 2024-04-24 at 5 49 25 PM

kingdonb:
We can consider reducing the footprint of the `#calendar` anchor even further on the landing page, and/or moving it under the Community heading.

This version already reduces the footprint of the landing page calendar substantially, reducing the number of events shown there, and successfully migrates a version of the calendar to the Community page under Project.

Screenshot 2024-04-24 at 7 23 57 PM

It could do better by duplicating less CSS, however, the look should be OK. All CSS has been tuned carefully. If you see anything else wrong with the look, please point it out and/or suggest an adjustment.

kingdonb:
I'm going to remove that "do not merge" commit, and come back with a preview link that is hosted off-netlify so we can continue to preview the site. This will break the deploy preview, but is necessary before merge, since `community` repo has its own set of changes that has to merge first + with this PR at approximately the same time.

kingdonb:
The deploy preview at https://docs-fluxcd-website-6lrhgsuf.fermyon.app/ is current

kingdonb:
Everything is in order for both PRs to merge now, whenever someone has time to review this 👍

stefanprodan (approved):
LGTM

Thanks @kingdonb 🏅

PR #1944, opened 2024-05-02
Update SOPS documentation for Azure to use workload identity

dipti-pai:
AAD Pod identity is deprecated in favor of workload identity. This PR 

- Adds documentation to use Azure workload identity to encrypt secrets using mozilla sops.
- Removes references to aad pod identity.


stefanprodan (approved):
LGTM

Thanks @dipti-pai 🏅

PR #1950, opened 2024-05-08
docs(helmrelease): typo

choeffer:


choeffer:
It could even make sense to rewrite the sentence to something like "If one or many of the `spec.chart.spec.valuesFiles` don't exist inside the chart, ... ", because I would expect, that it will fail in both cases, if one or many defined files are not present. 

Is this assumption correct? What do you think?

kingdonb:
Yes, they are applied in order, I would expect it to fail if any of the files listed are not present 👍

content/en/flux/guides/helmreleases.md
```diff
@@ -484,7 +484,7 @@ spec:
     replicaCount: 5
 ```
 
-If the `spec.chart.spec.valuesFiles` doesn't exist inside the chart, helm-controller will not be able to
+If one or many of the `spec.chart.spec.valuesFiles` don't exist inside the chart, helm-controller will not be able to
```
souleb, line 487: I think yo also need to consider `.spec.chart.spec.ignoreMissingValuesFiles`

PR #1963, opened 2024-05-13
Trigger specific branch deploy from GHA workflow

makkes:
That way we always trigger the build for https://fluxcd.io from this workflow.


kingdonb (approved):
I didn't know you could pass trigger branch as a parameter! This is much better, I can follow up with a PR to detect the latest minor branch and make it so you no longer need to hardcode a string there. But that can come later. Lgtm

PR #1965, opened 2024-05-13
Announce "Release 2.3" blog

kingdonb:
The Announce header on fluxcd.io main page can change to point at the new release blog, 🎉


stefanprodan (approved):
Thanks @kingdonb

PR #1967, opened 2024-05-14
Update dev guides to Flux v2.3

stefanprodan:
Update API versions in dev guides

### DIFF
--- a/content/en/flux/gitops-toolkit/packages.md
+++ b/content/en/flux/gitops-toolkit/packages.md
@@ -32,23 +32,24 @@ go get github.com/fluxcd/source-controller/api
 Import package
 
 ```go
-import sourcev1b2 "github.com/fluxcd/source-controller/api/v1beta2"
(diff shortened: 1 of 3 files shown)


content/en/flux/gitops-toolkit/packages.md
```diff
@@ -187,7 +188,7 @@ func main() {
   ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
   defer cancel()
 
-  // create a Helm repository pointing to Bitnami
+  // create a Helm repository pointing to https://kubernetes.github.io/ingress-nginx
```
souleb, line 191: why the change here? The repository points to bitnami still.

stefanprodan: Good catch, i wanted to change the repo but installing ingress-nginx in the default ns doesn't make sense. We should rewrite this with OCIRepo, but that's for another time.

souleb (approved):
LGTM!

PR #1972, opened 2024-05-15
feat(adopters): Adding Pets at Home

userbradley:
Adding Pets at Home as an adopter of Flux CD


stefanprodan (approved):
Thanks @userbradley 🥇

PR #1975, opened 2024-05-17
Add platformengineers to adopters and add improwised logo

payalraviya:
Add platformengineers to adopters


kingdonb (approved):
LGTM

Thanks @payalraviya for using Flux and Flagger, and for your support!

PR #460, opened 2021-08-13
Add Bink to adopters

cpressland:
Added Bink to Flux 2 adopters, added Bink logo to the Flux 1 section.

PR #1987, opened 2024-06-03
Remove references to Bink from adopters

cpressland:
Bink is no longer in business, just doing a bit of housekeeping as an ex-DevOps. Reverts https://github.com/fluxcd/website/pull/460


stefanprodan (approved):
Thanks @cpressland

PR #2077, opened 2024-10-30
Salt Lake City Kubecon updates

kingdonb:
This PR is ~WIP~ ready to go - with these:
* fluxcd/community#414
* fluxcd/community#413

(Will also need to add a CSS rule for the "qr-code-right-align" class) Not sure if any other changes are needed yet, (marking as a draft)

The order that these PRs should merge: first fluxcd/community#413 - then rebase fluxcd/community#414 and merge - finally merge this PR (and ensure the website build completes successfully!)

This one can also be backported for the public website


kingdonb:
Screenshots of how it looks:

Screenshot 2024-10-30 at 1 52 43 PM
Screenshot 2024-10-30 at 1 52 52 PM
Screenshot 2024-10-30 at 1 53 02 PM

If there was someone with a bit more web design chops, I'd say we should make it look nicer --

but this looks almost exactly like the source Google Page/Site website so I'm not going to mess with it, since it took a bit of doing to even get this to look like how that looks on the Flux web site!

In fact there's still a small unresolved issue in fluxcd/community#414 why we can't just merge this, I need to do some manipulating of the order of the elements in the Markdown file, in order to get things to play nice for HTMLs that float left and float right. (It turns out that HTML float left should be placed grammatically after the HTML float right. HTML considers the float left to be semantically first.

This is counter-intuitive to me and probably everyone who is not an HTML and CSS export, but it is an easy fix, and I can probably resolve it by moving things around on the source website, rather than adding more complexity to the script.

Almost ready for merge! Please have a look at the screenshots above and approve if you agree, I'll take care of whatever else needs to happen before this can finally merge. 👍 thank you!

kingdonb:
I am happy with the way this turned out,

* https://github.com/fluxcd/community/pull/414

Is marked ready for review, when that one merges, we should merge this one straight after as it contains CSS that is needed to render the next version of the `kubecon.md` page properly.

kingdonb:
(The deploy preview does not reflect the actual changes because the changes in fluxcd/community have yet to merge)

stefanprodan (approved):
LGTM

Thanks @kingdonb 🥇

PR #2078, opened 2024-10-31
Add guides to use Flux in the enterprise

nagyv:


nagyv:
@stefanprodan If you see value in having an additional guide on the topic, I'd like to ask you to take it over from here as I'm sure you have more insights than I do. Otherwise, feel free to close this PR.

kingdonb:
Thank you for contributing this! I will put something on the agenda for the next dev meeting, so we can talk about it.

PR #2086, opened 2024-11-07
Fix inconsistencies in receiver guide

kingdonb:
Fixes fluxcd/flux2#2240

And some other minor issues I found while running through the webhook receiver guide:

* A tab character at the end of a copy-paste block that fouls things up
* The name of the receiver service is misnamed above a copy paste block

### DIFF
--- a/content/en/flux/guides/webhook-receivers.md
+++ b/content/en/flux/guides/webhook-receivers.md
@@ -82,7 +82,7 @@ spec:
 
 Add any necessary annotations for your ingress controller and cert-manager to provide for encryption. Full configuration of cert-manager issuers, ingress controllers, and TLS are beyond the scope of this documentation.
 
-However, one common issue caused by Flux's default network policy securing the `flux-system` namespace is that unexpected traffic to any pod in that namespace is prevented by Flux's default network policy. The HTTP-01 ACME challenge ingress is blocked from receiving cert-manager traffic.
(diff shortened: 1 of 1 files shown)


kingdonb:
Let me also fix this one before we merge this:

* https://github.com/fluxcd/flux2/issues/2240

kingdonb:
This should be good to merge 👍

kingdonb:
There is some feedback about the networkpolicy:

* https://github.com/fluxcd/flux2/issues/2240#issuecomment-2462593628

Maybe should address this before merging. I am not very proficient in network policies so I'm not terribly surprised if I haven't done it perfectly.

kingdonb:
It sort of looks like cert-manager has its own helm values for network policy, which you can just `enabled: true` and get the behavior. But I tried that and it didn't get me anywhere. Maybe some adjustment is needed to accommodate the flux network policies in addition to those.

I will give this another try tomorrow, and if I can't figure it out, I'll remove the network policy stuff, or scale it way back.

(I personally would be happy with the sub-optimal network policy that allows anyone to communicate with the pods labeled http-01 - nobody is going into the flux-system namespace and launching arbitrary pods or relabeling pods unless a cluster is already thoroughly compromised, so I don't see that as a problem. But I am not sure I understand how network policy works so I'll do some testing before I put that in the docs.)

kingdonb:
The answer was a lot simpler than I thought. It even works on my "not such a good example" environment.

I don't know where we got the idea that a reverse policy was needed.

stefanprodan (approved):
LGTM

Thanks @kingdonb 🏅

PR #2091, opened 2024-11-12
Adding `--create-namespace` flag to the helm-install example

dwertent:


makkes (approved):
Thanks for this fix! 🫶

PR #2096, opened 2024-11-24
Add the Headlamp article as a resource

kingdonb:
There were a few articles hosted on weave.works but they are not accessible anymore, so had to be removed.

I am testing this feature to be sure I know how it works, since we do not have any live examples today!

(I'd like to add this one too:)
* https://medium.com/@mohamednoureldinn/fluxcd-azure-devops-oidc-authentication-916cc7bfe11f

### DIFF
--- a/data/resources.yaml
+++ b/data/resources.yaml
@@ -13,6 +13,11 @@ resources:
   # Note: To generate "thumbnail_url", visit https://embed.ly/docs/explore/extract
   # and enter the link to the resource. Once you do that, select any thumbnail url of choice to use.
 
+  - url: https://headlamp.dev/blog/2024/11/07/flux-ui/
+    title: "Headlamp Blog | From ClickOps to GitOps: A new Flux UI"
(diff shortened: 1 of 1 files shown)


kingdonb:
We need to add thumbnails and/or description text in order to use this feature effectively. I wonder if there is a streamlined way to do this.

data/resources.yaml
```diff
@@ -13,6 +13,11 @@ resources:
   # Note: To generate "thumbnail_url", visit https://embed.ly/docs/explore/extract
   # and enter the link to the resource. Once you do that, select any thumbnail url of choice to use.
 
+  - url: https://headlamp.dev/blog/2024/11/07/flux-ui/
+    title: "From ClickOps to GitOps: A new Flux UI"
```
stefanprodan, line 17: Can we add "Headlamp blog" or something that hints this is about Headlamp?

stefanprodan (approved):
LGTM

Thanks @kingdonb

PR #2097, opened 2024-11-25
Add FluxCD Azure DevOps article resource

kingdonb:
Testing the Resources / Article feature again. (also #2096)

Thanks Mohammed Nour for contributing this article about the new features in Flux 2.4 🎉 

(I asked if there was a branding image that goes with this blog, and it was suggested we just use the Flux logo as there isn't one.)

### DIFF
--- a/assets/scss/_index.scss
+++ b/assets/scss/_index.scss
@@ -88,10 +88,6 @@ main.td-main {
     margin: 0.85rem 0;
   }
 
-  h6.resource-desc {
-    padding: 0 14px 14px 14px;
-  }
-
   img {
     height: 220px;
   }
@@ -101,6 +97,9 @@ main.td-main {
   }
 }
 
+h6.resource-desc {
+  padding: 0 14px 14px 14px;
+}
 
 .td-content {
   div.lead  {
--- a/assets/scss/_index.scss
+++ b/assets/scss/_index.scss
@@ -98,7 +98,8 @@ main.td-main {
 }
 
 h6.resource-desc {
-  padding: 0 14px 14px 14px;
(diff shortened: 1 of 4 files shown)


kingdonb:
https://deploy-preview-2097--fluxcd.netlify.app/resources/

I think we should add a bit of description, since there is a place for it in the "article" format and the Flux logo doesn't fill the same 4x3 space as the others. Running this by the author to see what text we can put in there before we publish this.

stefanprodan:
The article is about migrating existing clusters from Azure PAT or SSH to OIDC. I think the title should say this.

kingdonb:
> In this article, I’ll guide you step-by-step through the process of setting up Flux source-controller and image-automation-controller to authenticate against Azure DevOps repositories using OIDC tokens

This sentence from the article I picked is probably too long for the description, and doesn't say exactly what you just said... but combining them, might be the right level of detail - how about this (let's see if it fits in the box):

> Migrating existing cluster Git sources and image automations from Azure PAT or SSH to OIDC

stefanprodan:
Maybe "Migrating AKS clusters to Azure DevOps OIDC-based auth"

kingdonb:
This LGTM

kingdonb:
Maybe better if it wasn't center-aligned, I'll tweak it...

kingdonb:
Did several iterations on the CSS and template for description in resources, I think this is better 👍

kingdonb:
Screenshot 2024-11-25 at 3 21 52 PM

Preview ^

stefanprodan (approved):
LGTM

Thanks @kingdonb 🥇

PR #2103, opened 2024-11-29
Weave dot works is down

### LINKED ISSUE #2102: weave.works domain no longer in use?
Apparently the domain has expired and now it is serving phisy looking content. Should links to `weave.works` be scrubbed from the website and other parts of the documentation?

kingdonb:
Fix #2102 

I was able to find a good copy of all of the Weaveworks blog content at Internet Archive, for every blog post or reference we've ever made there. Sadly, the registration for `weave.works` has lapsed - I don't think it'll come back without some kind of benefactor to rescue the domain at auction, but we can preserve blog references so that we are not serving up bad links.

If there are any references that we should simply remove instead of making into historical references to the archive, please point them out and I'll take another look. This first pass I merely checked if there was any reference before the Netlify account went into arrears - perhaps surprisingly, I don't think any of these articles were completely unrecoverable!


kingdonb:
Interestingly, the slack inviter can be replaced with a hard link to weave-community - I do not know if disclosing the invite code will make it invalid, but it was already disclosed by Internet Archive, so hopefully not.

There is still a small community there discussing tofu-controller and another talking about the next version of Weave Net, which has been resurrected as a [lead fork](https://github.com/rajch/weave#using-weave-net-on-kubernetes) with few changes, basically maintenance only, and another [smaller fork](https://github.com/kuttiproject/weave) with no Docker support; both are owned by @rajch

I was maintaining a live fork of the Weave GitOps docs, but with the loss of the domain, that one is down too. If anyone wants to take over hosting the Weave GitOps docs, or if you need to peruse the docs internally for any reason, they are of course all open source licensed; furthermore I've been able to pack them up as an OCI bundle for easy deployment on Kubernetes.

I'm using Fermyon's SpinKube for that. This was made possible by Spin 3.0 🙇

https://github.com/microscaler/gitops-docs/tree/spin-it-back-up

makkes (approved):
Excellent job!

PR #2130, opened 2025-01-23
Mark source controller's Github App Auth implementation as done in rroadmap

dipti-pai:
Mark source controller's Github App Auth implementation as done in roadmap


matheuscscp (approved):
Thanks @dipti-pai!

PR #2156, opened 2025-02-22
Add a note on Forgejo support

oligot:
Forgejo is a fork of Gitea (see their [Comparison with Gitea](https://forgejo.org/compare-to-gitea/)).

I found [this discussion on the main flux repo](https://github.com/fluxcd/flux2/discussions/4924) and wanted to know if the gitea boostrap command works with Forgejo. I've described what I've done in a blog post titled [Flux and Forgejo](https://oligot.be/posts/flux-forgejo/) if you are interested but in a nutshell, you can use the exact same commands with a Forgejo server and it works !

This adds a small note on the gitea boostrap page to let people know that they can use Forgejo instead of Gitea with Flux.


stefanprodan (approved):
LGTM

Thanks @oligot

PR #2165, opened 2025-03-05
Update website cron job for Flux 2.5

kingdonb:
I noticed after the Flux Bug Scrub last night, the website did not update this A.M.

It looks like we still have to update this workflow for each new minor/major release, to change the trigger branch.

### DIFF
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -12,4 +12,4 @@ jobs:
     steps:
       - name: Deploy current production branch on Netlify
         run: |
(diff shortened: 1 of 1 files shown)


kingdonb:
This is an improvement over the old workflow where we had to create a new secret! But it is still a manual step.

I will think about how to solve this, maybe we can get the latest release from GitHub API, parse it as semver, and do a string transformation to get the branch name in the form `vMAJOR-MINOR`.

We already have lots of Python and Bash code in the website pipeline.

matheuscscp (approved):

PR #2187, opened 2025-03-28
Add non-root install instructions

timonegk:
The installation script for the flux cli supports installation to a custom directory, this is now reflected in the documentation.


matheuscscp (approved):
Thanks @timonegk!

PR #2192, opened 2025-04-09
chore: add Beekeeper AG as adopter

sanadhis:
We (Beekeeper AG) uses FluxCD to manage our apps since ~3 years now 🚀


stefanprodan (approved):
Awesome! Thanks @sanadhis

PR #2201, opened 2025-04-16
docs(kustomize-controller): update feature-gates

zevisert:
I noticed today that the documentation for the kustomize-controller's feature-gates doesn't match what the controller has implemented: https://github.com/fluxcd/kustomize-controller/blob/60afb733273cbbaf8e53b206aa877a125d249233/internal/features/features.go#L23-L69

I was looking for `StrictPostBuildSubstitutions` and found that `DisableFailFastBehavior` was also missing from the docs


stefanprodan (approved):
LGTM

Thanks @zevisert

PR #2244, opened 2025-05-31
Add Flux Operator as an alternative bootstrap method

stefanprodan:
xref: https://github.com/fluxcd/flux2/pull/4749

### DIFF
--- a/content/en/flux/installation/_index.md
+++ b/content/en/flux/installation/_index.md
@@ -21,12 +21,12 @@ The Kubernetes cluster should match one of the following versions:
 
 | Kubernetes version | Minimum required |
 |--------------------|------------------|
-| `v1.29`            | `>= 1.29.0`      |
 | `v1.30`            | `>= 1.30.0`      |
-| `v1.31` and later  | `>= 1.31.0`      |
+| `v1.31`            | `>= 1.31.0`      |
+| `v1.32` and later  | `>= 1.32.0`      |
 
 {{% alert color="info" title="Kubernetes EOL" %}}
-Note that Flux may work on older versions of Kubernetes e.g. 1.25,
+Note that Flux may work on older versions of Kubernetes e.g. 1.29,
 but we don't recommend running [EOL versions](https://endoflife.date/kubernetes)
 in production nor do we offer support for these versions.
 {{% /alert %}}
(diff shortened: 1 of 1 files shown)


content/en/flux/installation/_index.md
```diff
@@ -143,7 +143,7 @@ Various configuration options are available at bootstrap time such as:
 
 * [Installing optional components](configuration/optional-components.md)
 * [Enforcing tenant isolation on shared clusters](configuration/multitenancy.md)
-* [Using workload identity on AWS, Azure and GCP](configuration/workload-identity.md)
+* [Using workload identity on AWS, Azure, and GCP](configuration/workload-identity.md)
```
matheuscscp, line 146: Comma before `and` is only used for independent sentences. In a list of items `and` already fulfills the purpose of a comma.

content/en/flux/installation/_index.md
```diff
@@ -157,16 +157,91 @@ The provider offers a Terraform resource called
...
+metadata:
+  name: flux
+  namespace: flux-system
+  annotations:
+    fluxcd.controlplane.io/reconcileEvery: "1h"
+    fluxcd.controlplane.io/reconcileArtifactEvery: "10m"
+    fluxcd.controlplane.io/reconcileTimeout: "5m"
+spec:
+  distribution:
+    version: "2.x" # automatic upgrade to the latest Flux version
```
matheuscscp, line 203: I'd prefer this example to be for the latest minor release patches instead of the latest minor releases.

matheuscscp (approved):

content/en/flux/installation/_index.md
stefanprodan: This would mean updating the example for every flux release and we’ll forget about it. Also this sample matches what’s in the operator readme.

matheuscscp: I'm fine including this as an item in the website release procedure, there's already a place where we do it anyway (the website home page).

I think we should update the sample in the operator readme.

We should definitely do more than our best to avoid breakages across minor releases, but sometimes it'll just slip through the cracks.

stefanprodan: People should read the docs and decide what upgrade policy to use. I see no reason why would we complicate the website release procedure for this. Also this is an ecosystem project, other maintainers shouldn’t have to do more work because of this docs addition.

stefanprodan: added

PR #2249, opened 2025-06-04
Fix GitHub Action syntax error and filter with if statement

felddy:
`create` events cannot be filtered with `branches`. `branches` be will silently ignored and match all create events.

See:
- https://docs.github.com/en/webhooks/webhook-events-and-payloads#create
- https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#create
- https://github.com/orgs/community/discussions/26286


stefanprodan (approved):
LGTM

Thanks @felddy 🥇

PR #2256, opened 2025-06-08
Add RBAC patch to remove Object-level Workload Identity

stefanprodan:
Add `Object-level Workload Identity RBAC` section to multi-tenancy docs with an example of how to disable the extra RBAC.

### DIFF
--- a/content/en/flux/installation/configuration/multitenancy.md
+++ b/content/en/flux/installation/configuration/multitenancy.md
@@ -157,3 +157,40 @@ patches:
       kind: ClusterRole
       name: "(flux-view|flux-edit)-flux-system"
 ```
+
+### Flux Object-level Workload Identity RBAC
+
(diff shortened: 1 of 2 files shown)


content/en/flux/installation/configuration/multitenancy.md
```diff
@@ -157,3 +157,40 @@ patches:
       kind: ClusterRole
       name: "(flux-view|flux-edit)-flux-system"
 ```
+
+### Flux Object-level Workload Identity RBAC
+
+Starting with v2.6, Flux supports object-level workload identity, which requires
+additional RBAC permissions to be granted to the controllers so that they can create Service Accounts tokens:
```
matheuscscp, line 164:
```suggestion
additional RBAC permissions to be granted to the controllers so that they can create ServiceAccount tokens:
```

matheuscscp (approved):

PR #2252, opened 2025-06-07
Add information about Freelens extension

dex4er:
Introduce the Freelens extension, a user interface for managing Kubernetes clusters, and update the documentation to include its details and visual representation.

PR #2258, opened 2025-06-08
Fix carousel, fix local build on MacOS

kingdonb:
I was able to fix the carousel

To get the build working at all on MacOS, first I had to:

* If a tool called "gsed" is available, use that one.
* If not, revert to "sed"

On MacOS sed is BSD sed and it fails around line 155, on the "Hack for fixing typo in the docs" regex

### DIFF
--- a/hack/import-flux2-assets.sh
+++ b/hack/import-flux2-assets.sh
@@ -9,6 +9,12 @@ if [ -z "${GITHUB_USER:-}" ]; then
     GITHUB_USER=fluxcdbot
 fi
 
+if [ "$(command -v gsed)" ]; then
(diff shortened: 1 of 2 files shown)


kingdonb:
Resolves the issue in #2252

kingdonb:
Can see it's fixed here:

https://deploy-preview-2258--fluxcd.netlify.app/ecosystem/#freelens

stefanprodan (approved):
LGTM

Thanks @kingdonb 🏅

kingdonb:
(this will get deployed to the website via a backport from #2259)

PR #2274, opened 2025-06-14
Update releases procedure, mention params.version

kingdonb:
I think there's a good reason why this string was not ever updated in the `hugo.yaml` of the main branch over the last 3 releases. I'm betting now that I have updated it there, we'll get backport failures from merge conflicts related to this line. Or the backports will try to override the version, and succeed, causing backport branches to inherit the wrong version.

This string is only ever used in the "archived_version" header. So it needs to be set in each backport branch, only.

Someone who has actually done releases can please validate - is this good information?


matheuscscp:
Thanks!

@kingdonb Please squash :pray:

PR #2320, opened 2025-08-15
Fix left sidebar to use full height (#2136)

### LINKED ISSUE #2136: CSS: Left sidebar doesn't utilize full height (and is confusing)
Screenshot:
Image

See lower left corner. 

The sidebar doesn't expand to use the whole area.

The current CSS, given the absence of scrollbars, gives the illusion that there is no more content, when there actually is.

sameersharmadev:
closes #2136 
image
If applied, this commit will ensure that the sidebar takes full available height while maintaining the scroll so that its obvious there is more content below

### DIFF
--- a/assets/scss/blocks/_nav.scss
+++ b/assets/scss/blocks/_nav.scss
@@ -83,4 +83,9 @@ nav.navbar {
     color: $gray-900;
     font-weight: unset;
   }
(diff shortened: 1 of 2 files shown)


matheuscscp (approved):
LGTM! 🚀

PR #2381, opened 2025-10-02
Pull API docs from release branch instead of release tag

matheuscscp:


stefanprodan (approved):
LGTM

Thanks @matheuscscp

PR #2383, opened 2025-10-03
Add Gateway API example to the webhook receiver guide

stefanprodan:
Update the docs to show how to expose notification-controller webhook receiver using Gateway API.


stefanprodan:
> it took me a second to realize a that service/webhook-receiver implicitly forwards 80->9292 in the last two examples

Good feedback, I've added this service and port mapping at the top!

matheuscscp (approved):
LGTM

PR #2394, opened 2025-10-14
fix(home): wrap tomorrow's zoom event URL for responsiveness

akshatsinha0:
Earlier in the mobile responsive versions of the https://fluxcd.io/ website, the zoom link was overflowing on the right that is why a x scroll was created:
image
But after wrapping it up and rendering the link:
image

I am not sure if this is required, so please let me know or is there any change I have to make along with it.


kingdonb:
Thanks for this @akshatsinha0

akshatsinha0:
> Thanks for this @akshatsinha0 

Maybe we should have cerntralised usage of color tokens, sizes (small, medium, large, etc), and other variables.

PR #2415, opened 2025-12-16
Add Causely to Adopters

svrnm:
Hi there, 

we are using FluxCD at Causely and are very happy with it, wrote a [small blog post about it today](https://www.causely.ai/blog/thank-you-fluxcd) and I thought it would also be time to have us added to the adopters list:-)

Thank you for maintaining this awesome project!


stefanprodan (approved):
LGTM

Thanks @svrnm awesome blog post 🏅

PR #2440, opened 2026-02-01
Add flux9s link to the Flux OSS ecosystem doc

dgunzy:
Signed-off-by: Daniel Guns <danbguns@gmail.com>

Hi @stefanprodan  @matheuscscp  - I would love to add `flux9s` as part of the Flux ecosystem on the site. Let me know if there is any other process I should follow, or anything else I should reference either in the codebase, or on my docs site to better align with Flux, and being part of the OSS ecosystem around it 😄


stefanprodan (approved):
LGTM 🚀

PR #2474, opened 2026-02-28
Thread of photos from FluxCon and KubeCon NA 2025

stealthybox:
Adding photos that can be used from our KubeCon community page


stefanprodan:
@stealthybox we should have some photos from last FluxCon with maintainers on stage. Let's add those too please.

PR #2503, opened 2026-04-04
Clean up outdated Flux adopters and references

parisnakitakejser:
During review of external “Flux adopters” and reference links, we found that many sources are no longer valid or accurate. Several links:

**No longer resolve (dead URLs)**
- Flux v2
- Flagger
- Flux v1 (legacy)

This makes the current list unreliable for documentation.


stefanprodan (approved):
LGTM

Thanks @parisnakitakejser 🏅

Issue #1487, opened 2023-05-03
/flux does not follow documentation style guide


dholbach:
The documentation (https://fluxcd.io/flux/) does not follow the [Documentation Style Guide](https://fluxcd.io/contributing/docs/style-guide/) at numerous instances. For example,

- Reconciliation [here](https://fluxcd.io/flux/concepts/#reconciliation) being a new term that is introduced is not italicized ([as per](https://fluxcd.io/contributing/docs/style-guide/#use-italics-to-define-or-introduce-new-terms)) .
- The second point in the [Deploy podinfo application](https://fluxcd.io/flux/get-started/#deploy-podinfo-application) talks about how the structure of `fleet-infra` repo be. It then proceeds to show the folder structure, without telling the user how to generate it, assuming a specific level of understanding.

PR #2528, opened 2026-04-21
Italicize new term introductions in core concepts

karthi-palanisamy:
Per the Flux documentation style guide, new terms should be italicized when first introduced. This applies the convention to the remaining terms in /flux/concepts: GitOps, Gitless GitOps, GitOps Toolkit, bootstrap, Continuous Delivery, Continuous Deployment, and Progressive Delivery.

Refs #1487


matheuscscp (approved):
LGTM

PR #2536, opened 2026-04-23
Add Hetzner to adopters

marvin-roesch:
We're happily using Flux across the company and would love to show we're part of the community.


stefanprodan (approved):
Thanks @marvin-roesch 🏅

PR #2594, opened 2026-06-29
Import controller options from controller docs

### LINKED ISSUE #5959: Move controller flags and feature gates to controller repos
Today we document flags and gates in the website repo, which is super annoying, makes us have to open 2 PRs instead of one.

We should move those things to the controller repos and make the website build script pull them from the controller repos just like we do it for the CRD API docs.

matheuscscp:
Part of https://github.com/fluxcd/flux2/issues/5942

Closes: https://github.com/fluxcd/flux2/issues/5959


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

PR #2612, opened 2026-07-08
Fix broken bitnami-labs references to bitnami

jensnieulandt:
The bitnami-labs repository has moved, therefore some links return a 404. Using the labs repo causes the sealed-secrets HelmRelease to fail.

Announced here: https://github.com/bitnami/sealed-secrets/issues/1982


stefanprodan (approved):
LGTM

Thanks @jensnieulandt

PR #2616, opened 2026-07-12
Correcting the location to `interval` value

ringods:
I copied [this example](https://fluxcd.io/flux/installation/upgrade/#upgrade-with-flux-operator) to my repo, but it failed with this error:

Screenshot 2026-07-12 at 19 34 33

There is no value at `inputs.provider.interval`. It should be just `inputs.interval`. After my fix, it works correctly.


matheuscscp (approved):
LGTM

PR #2626, opened 2026-07-24
Update references to CNCF calendar

kingdonb:
The old cncf-flux-dev calendar is supposed to go away:

https://lists.cncf.io/g/cncf-flux-dev

The new Flux calendar lives on LFX Platform and actually requires login *even less* than the old cncf-flux-dev did, anyone can subscribe totally anonymously as far as I can tell. The docs update reflects the difference.

There is a parallel PR in the community repo that I will link back here when I post it, momentarily.

### DIFF
--- a/hack/import-calendar.py
+++ b/hack/import-calendar.py
@@ -27,7 +27,7 @@ import recurring_ical_events
 import urllib3
 import yaml
 
-CAL_URL = 'https://lists.cncf.io/g/cncf-flux-dev/ics/9524119/1081862612/feed.ics'
+CAL_URL = 'https://webcal.prod.itx.linuxfoundation.org/lfx/a092M00001IfTjDQAV'
 
 TOP_LEVEL_DIR = os.path.realpath(
     os.path.join(os.path.dirname(__file__), '..'))


kingdonb:
For admins the login to the new calendar and "PCC" Project dashboard is via Openprofile.dev

First log into your LFX account, then find "Tools" then Project Control Center

These references were provided by the CNCF for us to manage the new system and learn its capabilities, in brief:

* https://www.youtube.com/watch?v=OKNR_QHGXgM
* https://www.youtube.com/watch?v=KOSRy4c4Yv8

I'm kicking the tires right now. Since next week's meeting is our last meeting on the old Enterprise Zoom account platform, we're going to try to adjust the schedule and use the new platform one week ahead of the deadline.

When we merge these two PRs, the website should automatically update to use the new calendar source. It is possible that the preview will not show this correctly. (Let's find out...)

kingdonb:
Wow, that actually worked:

Screenshot 2026-07-23 at 9 27 36 PM

~~All indications are, these links include the Zoom login code and don't require password (like we have always had, as we prefer) - this LGTM right now but it's late for me, and I need a second review before it can merge anyway.~~

kingdonb:
LGTM with:

* https://github.com/fluxcd/community/pull/474

kingdonb:
As I am reviewing the Zoom link and exactly how it works, I think there is a possibility that you will indeed need an LFX Platform account to get into meetings from now on. Even as a guest. But let's hope they've made it unobtrusive enough that it will not be an obstacle for us. Edit: You can use any of Google, GitHub, Linkedin, or an email address to register for an LFX account instantly. Let's test it out on Wednesday next week, and see if it can work for us.

Meanwhile, we can merge this ahead.

kingdonb:
I think the calendar entries are all correct now. I left the "old" Zoom link intact for next Wednesday's meeting as we discussed, but added the new Zoom links to the old cncf-flux-dev calendar entries, so it should be seamless for anyone that hasn't updated their calendars.

While I think the CNCF Zoom enterprise accounts are closing and that has been well-advertised, I don't think there's any reason to believe the old cncf-flux-dev mailing list will not survive the transition. Even still, I've added a note to the end of calendar description on each meeting so that folks can find the new calendar, in the event that cncf-flux-dev lists site is also on the block and I just did not read the memo.

Whether community members subscribe the old way via lists.cncf.io or the new way via https://zoom-lfx.platform.linuxfoundation.org/meetings/fluxcd?view=week they will get to the correct meeting and only have to deal with the LFX Platform login once.

stefanprodan (approved):
LGTM

Thanks @kingdonb

PR #2632, opened 2026-07-30
Add 24metrics to adopters

migueltanada:
Adding 24metrics to the Flux v2 adopters list.

24metrics is an ad fraud detection and traffic quality platform. We help affiliate networks, advertisers and agencies detect click fraud, screen partner traffic and enforce compliance in real time.

We have been running Flux in production for about 5 years. Today it manages GitOps across all our staging and production Kubernetes clusters — everything from cluster infrastructure to our application workloads, with Flux image automation handling continuous delivery of our services.

Flux has been rock solid for us at this scale, and we're happy to be listed.


stefanprodan (approved):
LGTM

Thanks for using Flux!

PR #2634, opened 2026-08-04
Add blog post on config change magnitude checks

1991santhu:
This post develops the request in [fluxcd/flux2#5512](https://github.com/fluxcd/flux2/issues/5512) — configurable safeguards against misconfiguration — which has been open since September 2025 without maintainer input.

The argument: config pipelines validate whether a change is well-formed and whether it is permitted, and never ask whether its *effect* is plausible. Pruning to nothing is convergence working exactly as designed; what is missing is any notion of magnitude. The post credits #5512 up front for both the "100 objects down to 1" heuristic and the idea of signalling an expected decrease out of band, and sets out to generalise them.

What it adds:

- A working CI guardrail built on `flux diff kustomization`, using its documented exit codes and counting the `deleted` markers. I ran this against a live cluster (kind + `flux install`, podinfo Kustomization with pruning enabled) rather than writing it from the docs.
- Two honest caveats: parsing human-readable CLI output is brittle, which is an argument for a structured output mode on `flux diff`; and CI checks the cluster as it was when CI ran, so the removal count approved on a long-lived PR is not necessarily the one applied. That time-of-check/time-of-use gap is why the check ultimately belongs in the reconciler.
- Existing tooling acknowledged — [`flux-local`](https://github.com/allenporter/flux-local), which was recommended in the issue thread, already covers the diff-in-CI half.
- Context from neighbouring ecosystems: Terraform/OpenTofu users routinely evaluate a rendered plan with OPA, and the OpAMP spec has the same missing rollout semantics.

Happy to take direction on framing, length, or whether this fits the blog at all. If there is appetite for the design, I would like to write it up as an RFC against #5512.

Preview note: dated today so it renders in the Netlify preview.

### DIFF
--- a/content/en/blog/2026-08-03-how-large-is-this-change/index.md
+++ b/content/en/blog/2026-08-03-how-large-is-this-change/index.md
@@ -1,0 +1,235 @@
+---
+author: Santhosh Kumar Somarapu
+date: 2026-08-03 12:00:00+00:00
+title: "How large is this change? The question a GitOps pipeline never asks"
+description: "Flux will gladly tell you whether a change is well-formed. If a typo tells it to delete your entire production namespace, it will not ask twice."
+url: /blog/2026/08/how-large-is-this-change/
+tags: [ecosystem]
+---
+
+There is an [issue open in the Flux tracker](https://github.com/fluxcd/flux2/issues/5512) that
+describes a bad afternoon. Its author pointed a `FluxInstance` at a directory that turned out to be
+empty, and Flux, in their words, "gleefully proceeded to delete everything it was managing before."
+
+Read the mechanics and nothing malfunctioned. Git served the source. The Kustomization built. The
+diff was computed correctly against live state. The apply succeeded, resource by resource, exactly
+as instructed. There is no component in that chain you would file a bug against.
+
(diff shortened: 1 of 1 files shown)


stefanprodan:
@1991santhu I suggest posting this in [Discussion](https://github.com/fluxcd/flux2/discussions).

Also flux-local has been deprecated, there is a successor that I didn't had time to look into it: https://github.com/home-operations/konflate

1991santhu:
Thanks @stefanprodan, that's a fair steer — I'll move it to Discussions.

And thanks for the flux-local catch. I checked and its repo now says the same: deprecated and archived, pointing at `flate` for offline validation/rendering and `konflate` for PR review. I'll correct that reference — the point I was making there was "you may not need to build this yourself", and konflate is the right thing to send people to.

Happy to close this PR. If the discussion goes anywhere useful, I'd still be glad to write it up properly for the blog later, or as an RFC against fluxcd/flux2#5512 — whichever is more useful to you.

1991santhu:
Closing per @stefanprodan's steer — moved to a discussion: https://github.com/fluxcd/flux2/discussions/6025 (with the flux-local reference corrected to flate/konflate).

PR #2635, opened 2026-08-05
Update roadmap for 2.9

matheuscscp:


stefanprodan:
Can you please add Zoom and Incident io NC providers to the 2.10 roadmap, they are merged.

stefanprodan (approved):

Issue #2571, opened 2026-06-11
Summer Blog Series :))


stealthybox:
We've got a lot going on across the project, and we've got to make sure people know about it! haha

We're thinking 6 blog posts

## Plan:
- [x] Flux 2.9 release (@stefanprodan)
  -  June 30 #2596
- [x] Flux 10 Year Anniversary! (@stealthybox)
  -  July 7 #2610
- [x] Flux Schema (@stefanprodan)
  -  July 13 #2604
- [x] OpenBao Integrations (@matheuscscp, @datosh, @cipherboy, @stealthybox)
  -  July 27 #2619
- [x] Field Ignore Rules (@dipti-pai)
  -  Aug 3 #2625
- [x] Flux Mirror (@stealthybox)
  -  Aug 17 #2641 
- [ ] flux9s (@dgunzy)
  -  Aug 24  tbd #2642 
- [ ] New Receiver Workflows (@matheuscscp)
  -  To be defined

### Flux 2.9 release
we need our traditional post to cover everything in the release.
we can have a subnote pointing people to use flux-mirror for their HelmChart workflows, since HelmRepo/HelmChart are showing their age, and flux-mirror can act as both an OCI and HTTP chart supply chain diode -- note can say "blog post to come!".
Notably, we have some big features that won't be getting posts immediately:
- SSH Key Verification for Git
- managedFields Ignore Rules
- Private Sigstore Stack
- Solidified Helm 4 Support
- more
 
### Flux 10 Year Anniversary!
Our beloved project is turning 10. This is a massive symbol of everything all of maintainers, contributors, end-users, and community members have poured into the project over the years and we should look back on it fondly, knowing we have ,unquestionably, another decade in front of us.

### Flux Mirror
Flux mirror primary use-case is something everyone needs:
  - unix'y tool that does one thing well which is to move the bits of your OCI images, Helm Charts, Desired State, and any other OCI artifacts such as kyverno policies or machine images (bootc) into your OCI registry of choice
  - We can mention that it's a great replacement for chartsyncer for those who are worried about the license change
  - We should highlight that it's a supply chain focused tool built on top of one of the most trustworthy infra supply chains in OSS
  - SLSA build provenance and all that
These are using the new Plugin system :)

### Flux Schema
Another really solid tool shipped as a plugin that fits into your GitOps workflows --
[...]

PR #2641, opened 2026-08-12
Blog: Introducing Flux Mirror

stealthybox:
In addition to basic functionality, I included some ecosystem and historical context for why it's important for people to mirror their own dependencies. I also talk about implementing a "Supply Chain Diode" with minimum age as an example policy.

https://github.com/fluxcd/website/issues/2571

commits:
- **blog: Flux Mirror**
- **Add Flux cake to 10yr blog :)**

PR preview link
https://deploy-preview-2641--fluxcd.netlify.app/blog/2026/08/flux-mirror/


content/en/blog/2026-07-22-flux-mirror/index.md
```diff
@@ -0,0 +1,343 @@
...
+permissions:
+  contents: read
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: fluxcd/flux-mirror/actions/setup@main
+      - run: flux-mirror sync flux-mirror.yaml --no-progress
```
stefanprodan: Please add docker/login action in here, sync can not work without a write token. You could copy the whole yaml from https://github.com/fluxcd/flux-mirror/tree/main/actions/setup#usage

stefanprodan (approved):
LGTM 🚀 

Please squash the last commit.

matheuscscp:
I don't think this post is the right place for an official communication about putting HelmRepo and HelmChart in maintenance mode, but I wonder if we can use this post and be a bit more vocal and opinionated about it... We say this across GH issues but there's nothing strong in docs anywhere about this, so we often hear "interesting, I didn't find anything about this in any docs". The use case that was the spark to create flux-mirror was mirroring Helm HTTP to OCI, so I think it should have a lil more weight on the post

stealthybox:
@matheuscscp great point.  I've added this section 07c25f039118ae89c96c70b55ff113772dbd914d
Kube-prometheus-stack uses the prometheus-community index which is pretty big (like 6MB), but they publish OCI charts for everything now.

I did find that OpenBao publishes OCI charts but Vault does not, and the Hashi projects share an index that's like 384KB or something, so I used that as a relatable example.

content/en/blog/2026-07-22-flux-mirror/index.md
```diff
@@ -0,0 +1,424 @@
...
+with every release, and because of the way the Helm SDK works, every client
+has to download and parse the whole index just to resolve a single chart version.
+These indexes can vary in size from 50KB to a few megabytes.
+The frozen bitnami-legacy index is actually 27MB! Internal HTTP repos within
+your own company's Artifactory can have this exact same performance issue.
+
+OCI Helm charts don't have an index. Each chart resolves by name and tag like any
+other artifact, which is why `OCIRepository` sources perform significantly better
+in source-controller than HTTP `HelmRepository` sources. Big community repos like
+prometheus-community, grafana, and ingress-nginx have started publishing OCI
```
stefanprodan: ingress-nginx is dead, let's not mention it

matheuscscp (approved):
Thanks @stealthybox! Great addition!

LGTM! 🚀

content/en/blog/2026-07-22-flux-mirror/index.md
stealthybox: swapped to cert-manager

PR #2647, opened 2026-08-18
Add kdvz to adopters

sb-kdvz:


stefanprodan (approved):
LGTM

Nice to see a german public IT service provider using Flux 🏅

